PR #566, opened 2014-12-05
Exclude non-local service and national numbers from geocoding in Finland

anssih:
This is a copy of https://codereview.appspot.com/174570043/ . Since there was no activity there and libphonenumber has now moved to github, here is a PR instead.

Per http://www.viestintavirasto.fi/attachments/maaraykset/Maarays_32.pdf
(official numbering plan), in Finland (+358):

- 020*: National service and subscriber numbers
- 029*: National subscriber numbers
- 030*: National service and subscriber numbers
- 039*: National subscriber numbers
- 060*: National service numbers
- 0800*: National toll-free service numbers

The local subscriber numbers starting with 02,03,06,08 have 1..8 as
their third number.

However, libphonenumber assigns 02*, 03*, 06* and 08* to specific
locations in their entirety, causing e.g. some national service numbers
to be incorrectly geocoded.

Instead of using catch-all geocoding for 02,03,06,08, add the actually
local subscriber telephone number allocations 0X[1-8] separately.

### DIFF
--- a/resources/geocoding/en/358.txt
+++ b/resources/geocoding/en/358.txt
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 # Generated from:
+# http://www.viestintavirasto.fi/attachments/maaraykset/Regulation_32.pdf [2015-04-15]
 # http://en.wikipedia.org/wiki/+358 [437010438]
-# Expanded 3585 prefix to exclude mobile.
+# Some prefixes expanded to exclude mobile/nationwide numbers.
 
 35813|North Karelia
 35814|Central Finland
@@ -23,8 +24,22 @@
 35817|Kuopio
 35818|Åland Islands
 35819|Nylandia
-3582|Turku/Pori
-3583|Tavastia
+35821|Turku/Pori
+35822|Turku/Pori
+35823|Turku/Pori
+35824|Turku/Pori
+35825|Turku/Pori
+35826|Turku/Pori
+35827|Turku/Pori
+35828|Turku/Pori
+35831|Tavastia
+35832|Tavastia
+35833|Tavastia
+35834|Tavastia
+35835|Tavastia
+35836|Tavastia
+35837|Tavastia
+35838|Tavastia
 35851|Kymi
 35852|Kymi
 35853|Kymi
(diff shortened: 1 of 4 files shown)


anssih:
Here is an English version of the official numbering plan referenced above:

https://www.viestintavirasto.fi/attachments/maaraykset/Regulation_32.pdf

keghani:
Thanks Anssi!

Please point me to the page numbers in the English document where I may verify the ranges you outlined above.

And please digitally sign the the contributor agreement here: https://cla.developers.google.com/clas

Then I'll review your patch, followed by #664.

anssih:
Page 19: 020* and 029* are nationwide service/subscriber numbers
Page 20: 030* and 039* are nationwide service/subscriber numbers
Page 21: Some 060* ranges are nationwide service numbers, rest of 060* and 069* reserved
Page 24: 0800* are nationwide toll-free numbers, rest of 080* and 089* reserved.

Section 6, paragraph 3, on page 3 clarifies that "nationwide" numbers are non-geographic.

The document apparently uses slightly different nomenclature - I used "national" in the commit where the document says "nationwide", while the document uses "national" to mean any domestic number...

I've already signed the CLA, the page says:
Google Individual CLA 	Anssi Hannula 	May 01, 2014 03:59 PDT
I hadn't filled in the "github username" field that was added since then, though, but I did that now as well.

resources/geocoding/en/358.txt
```diff
@@ -13,8 +13,9 @@
 # limitations under the License.
 
 # Generated from:
+# http://www.viestintavirasto.fi/attachments/maaraykset/Maarays_32.pdf [2014-11-23]
```
keghani: Please change to
  http://www.viestintavirasto.fi/attachments/maaraykset/Regulation_32.pdf [2015-04-15]
as that's what I'm using to verify and we can both agree on that source.

resources/geocoding/en/358.txt
```diff
@@ -13,8 +13,9 @@
...
 # http://en.wikipedia.org/wiki/+358 [437010438]
-# Expanded 3585 prefix to exclude mobile.
+# Some prefixes expanded to exclude mobile/service/national numbers.
```
keghani: Change "mobile/service/national" to "mobile/nationwide" to minimize confusion between this and document.

resources/geocoding/en/358.txt
```diff
@@ -33,6 +48,20 @@
...
+35868|Vaasa
+35881|Oulu
+35882|Oulu
+35883|Oulu
+35884|Oulu
+35885|Oulu
+35886|Oulu
+35887|Oulu
+35888|Oulu
 3589|Helsinki
```
keghani: According to page 24 this should also be expanded to include 35891 through 35898.

anssih: OK, didn't do that since 90 and 99 aren't in current use, but you are right as that might change in the future...

anssih:
Updated commit according to comments, thanks.

PR #571, opened 2014-12-14
Rpm packaging

dimbleby:
These changes allow me to build both 32-bit and 64-bit RPMs for my Centos box.  They comprise:
-  fixing up 32-bit build warnings
-  make sure that 64-bit libraries go in lib64 (per #519)
-  introduction of CMake variable 32BIT, to allow user to force 32-bit compile
-  introduction of CMake variable BUILD_STATIC_LIB, to control whether or not to bother building static libraries (I don't want them, but I've left default as ON since that's what the current behaviour is)
-  use CPack to build the rpm


dimbleby:
I'm using 32BIT as something that's either defined or not defined.  I promise that what I've done works fine, but if you're keen on some alternative formulation please say so and I'll tweak.

I'll stick in some spaces per your second comment and squash that change into this commit.

hagbard:
This appears to have broken our internal build, so I am going to have to revert it.

dimbleby:
Well that's unfortunate!

Obviously I'd be keen to keep or fix this rather than remove it altogether - anything I can do to help?

dimbleby:
Oh I see, I'd written `$BUILD_STATIC_LIB` in two places where I should have had `${BUILD_STATIC_LIB}`.

I build with this variable set to OFF - which is why I failed to spot the problem.  Apologies, this was careless.

Any chance we can get this pull request reinstated, with that fix?  Let me know if you'd like a fresh pull request from me.

PR #572, opened 2014-12-14
Clean compile with CMAKE_BUILD_TYPE=RELEASE

dimbleby:
Fix unused-variable warnings when building with CMAKE_BUILD_TYPE=RELEASE


dimbleby:
I've tweaked per comments from @pliard-chromium.

PR #577, opened 2014-12-24
Add option to load the config resources through a different class loader

askannon:
This can be useful since prefixmapper and geocoder are packaged in two jars and in certain environments may be on different class paths (OSGI madness, etc.) and thus can result in a NPE in PrefixFileReader.java:56.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/577)


keghani:
Hi @askannon,

Is this still an issue after #1300?

askannon:
Hey @keghani,

I am afraid it doesn't solve the issue. While adding OSGI manifest information makes it easier for general class loading it doesn't solve the issue of file loading between artifacts.

In this specific case the PhoneNumberOfflineGeocoder from the "geocoder" JAR is using the PrefixFileReader from the "prefixmapper" JAR to load a file from the "/com/google/i18n/phonenumbers/geocoding/data/" directory in the "geocoder" JAR.
But since PrefixFileReader is using PrefixFileReader.class.getResourceAsStream() load the file it is using the classloader for the "prefixmapper" JAR and can't find the files from the "geocoder" JAR classloader.

keghani:
Thanks for clarifying @askannon! 
We'll look into this, but to set expectations, metadata updates are the priority for us to ensure the data is fresh.

keghani:
Sorry this has been sitting here a while!
It's better tracked as an issue since we'll need to change it in several places. I'll close and track in #1757. Thanks for taking a stab at the PR!

PR #582, opened 2015-01-12
Move abkhazian phonenumbers into separate territory block

kaineer:


dlyongemallo:
The organisation of the metadata file must follow the administrative divisions used by the ITU:
http://www.itu.int/oth/T0202.aspx?parent=T0202

PR #584, opened 2015-01-16
Configure pom.xml for Github

roes:


java/pom.xml
```diff
@@ -31,9 +31,9 @@
   </licenses>
 
   <scm>
-    <connection>scm:svn:http://libphonenumber.googlecode.com/svn/trunk/java/</connection>
-    <developerConnection>scm:svn:https://libphonenumber.googlecode.com/svn/trunk/java/</developerConnection>
-    <url>scm:svn:http://libphonenumber.googlecode.com/svn/trunk/java/</url>
+    <connection>scm:git:https://github.com/googlei18n/libphonenumber.git</connection>
```
andyst: I think we need the same value for connection and developerConnection.

http://maven.apache.org/pom.html#SCM explains that the former needs read, the latter write access. There's no reason to use a different value / format for the two as far as I can see.

Most of the examples I found use the scm:git:git@github.com:<project>.git format

E.g.
https://github.com/sonatype/sonatype-aether/blob/master/pom.xml
http://blog.sonatype.com/2009/09/maven-tips-and-tricks-using-github/#.VLlDpWTF-iJ
http://stackoverflow.com/questions/20581427/using-maven-release-plugin-with-github-you-cant-push-to-git-github-com

andyst:
Thanks for the change! Just a nit. Let's try to run this on Monday.

PR #587, opened 2015-01-20
In JavaScript, strings are not errors!

EgoAleSum:
See http://www.devthought.com/2011/12/22/a-string-is-not-an-error/

PS: I also fixed URLs in the README file


lararennie:
Superseded by https://github.com/googlei18n/libphonenumber/pull/1459

PR #595, opened 2015-01-31
Add node port (npm package) to README

ruimarinho:
I've added the node port which is basically an npm package wrapper for the javascript version of this library. It is very convenient for node.js developers to be able to just require this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,3 +110,4 @@ Several people are porting the phone number library to other languages. Here are
   * Python: https://github.com/daviddrysdale/python-phonenumbers
   * Ruby: https://github.com/sstephenson/global_phone
   * PHP: https://github.com/giggsey/libphonenumber-for-php
+  * JavaScript (node): https://github.com/seegno/google-libphonenumber


andyst:
@jias0001 thoughts?

My $0.02:
 - libphonenumber has a JavaScript version. If we promote another JavaScript version in the README, then it should be for good reasons. Also, if there are other JavaScript forks (and not just this one), we should probably mention those too.
 - We should look at the differences between @ruimarinho's port/fork and the original JavaScript version. Which of the reasons / differences can and should be integrated into libphonenumber? Which differences are worth keeping separate?
 - If we mention the fork in the README, then the label should be different. Instead of "JavaScript (node)" maybe "Node.js"?

ruimarinho:
Hoping I can clarify some of those points:

* Requiring `libphonenumber` from node is not a straightforward task. It depends on google closure's library in order for `goog` di to work.
* The *only* difference between the original JavaScript version and the one on this repo is that errors that are thrown as strings are converted to proper `Error` instances. The API 
* Sure, it looks better! Maybe add npm to help clarify the intention?

Thanks for the feedback.

ruimarinho:
@andyst I've just released [version 0.2.0](https://www.npmjs.com/package/google-libphonenumber) with `libphonenumber@7.0.3`. The only difference from the vanilla JavaScript version provided by `libphonenumber` is entirely on the [main file](https://github.com/seegno/google-libphonenumber/blob/master/index.js) - and that's it.

jias0001:
Sorry guys, somehow I missed the emails on Feb 2. Just saw Rui's reply
today.

Re: throwing Error instead of string - I remember sometime last year we
have updated our code to mostly throwing Error instead of string. Please
feel free to send us a patch to update the rest if they are missed from
that update.

Re: Readme - we try to keep the list as short as possible. For the known
ports, we try to only list one per language (the one best maintained). Also
note that we don't tend to list wrappers in this file, as there can be very
many of them.

To sum up, my opinion here is that we probably won't accept this patch to
list the wrapper in our README, but please do send other patches to improve
the lib, which makes it easier for you to maintain your wrapper downstream.
Does this make sense?

ruimarinho:
@jias0001 regarding throwing errors, unfortunately it still lacks throwing proper errors on multiple places (see [example](https://github.com/googlei18n/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js#L2965)). I'll see if I can contribute patches to make it more consistent though.

Your explanation makes sense but I think there is a [real demand](https://www.npmjs.com/search?q=libphonenumber) for a proper way to use this library on node.js. If this is something that can be done directly from javascript version, then I'm all for having it on a wiki or any other place of documentation. However, sourcing the closure library on node.js is not just a `require` away. Because of this, I consider this "wrapper" to actually be a port for node.js, even if they share the same language.

I would certainly prefer if this package could be officially distributed via `npm` without resorting to any modifications but, until then, I think the community will continue to search for a version that they can simply do `require('libphonenumber')` and it just works like every other node module.

PR #601, opened 2015-02-06
Update demo.html

jaffadog:
google closure has moved to github


andyst:
The next release will include a similar fix (making the URL relative as in the JavaScript tests).

PR #602, opened 2015-02-07
Fix brazilian ddds.

jasperes:
Filter Brazilian DDD's, blocking invalids DDD's.

### DIFF
--- a/resources/PhoneNumberMetadata.xml
+++ b/resources/PhoneNumberMetadata.xml
@@ -3109,7 +3109,15 @@
         <numberFormat nationalPrefixFormattingRule="($FG)"
             pattern="(\d{2})(\d{4})(\d{4})"
             carrierCodeFormattingRule="$NP $CC ($FG)">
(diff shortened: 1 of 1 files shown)


jasperes:
Font: http://www.embratel.com.br/Embratel02/files/secao/06/13/365/codigos_ddd.doc

keghani:
Thanks for this PR @jasperes! Sorry we missed it at the time.

This is working as intended.

We don't want more specific leading digits for the "(\d{2})(\d{4})(\d{4})" formatting pattern in Brazil, [1-9][1-9] is intentional. The leading digits are only used for formatting numbers, not to determine validity. The <fixedLine> pattern is more specific, and that's what's used for validity.

You can check the demo to verify: http://libphonenumber.appspot.com/phonenumberparser?number=23+0000+0000&country=BR
23 0000 0000 is invalid but possible. It gets formatted as (23) 0000-0000 in national format.

Issue #592, opened 2015-01-24
AsYouTypeFormatter may cause an exception in CN


Mortimer2013:
In fact, I encountered this issue on Android-L VM. But you can still reproduce it with this code:

	public static void main(String[] args) {
		String countryISO = "CN";
		PhoneNumberUtil pnu =PhoneNumberUtil.getInstance();
		AsYouTypeFormatter formatter = pnu.getAsYouTypeFormatter(countryISO);
		
		String phoneNumber = "179513801031234";
		for (int i = 0; i < phoneNumber.length(); i++) {
			System.out.println(formatter.inputDigit(phoneNumber.charAt(i)));
		}
	}
Thank you for your reply! :)

![screenshot](https://cloud.githubusercontent.com/assets/3844627/5886168/0cc73d7e-a3c8-11e4-9f9d-fcf926a5a779.png)

dlyongemallo:
Thanks for reporting this! This will be fixed in the next release.

PR #614, opened 2015-02-18
Fix bug in AsYouTypeFormatter where we hit a IndexOutOfBoundsException

andyst:
in its Java implementation when extracting the Chinese national prefix 17951 and didn't reset the format template / lastMatchPosition.

Includes corresponding changes for C++ and JavaScript as well to keep the implementations in sync. (C++ and JS didn't exhibit buggy behavior because the corresponding substring methods don't throw errors for invalid start positions.)

Bugfix for https://github.com/googlei18n/libphonenumber/issues/592


dlyongemallo:
I think the comment added to the Java file should analogously be made in the C++ and js files, to keep them in sync.

Issue #611, opened 2015-02-14
Argentina - Invalid Format for Area Codes 3487 and 3489


mdaloia:
Given this valid E.164 number: +54 3487 44 1111 (the last part was modified to preserve the identity)
**Current National Format:** 0348 744-1111
**Expected National Format:** 03487 44-1111

Given this valid E.164 number: +54 3489 48 1111 (the last part was modified to preserve the identity)
**Current National Format:** 0348 948-1111
**Expected National Format:** 03489 48-1111

**Version used:** 7.0.2

This information is taken from the Argentinian bureau of communications called CNC. You can download this Excel where are all the assigned Area Codes (column INDICATIVO) and the Block number (column BLOQUE, which in this case are 44 and 481 respectively):

http://www.cnc.gob.ar/multimedia/noticias/archivos/201411/archivo_20141118022303_914.zip
(The link was obtained from this URL: http://www.cnc.gob.ar/asignaciones-a-la-fecha_p445 (On the right side there is a link with a "Númeración Geográfica" as a label))

You can filter INDICATIVO by 3487 and 3489, and BLOQUE by 44 and 481 to see all of them.

Below there is a Workaround that we made in the PhoneNumberMetadata.xml to change this behaviour:

We modified these 2 lines (872 and 940): https://github.com/googlei18n/libphonenumber/blob/libphonenumber-7.0.2/resources/PhoneNumberMetadata.xml#L872
https://github.com/googlei18n/libphonenumber/blob/libphonenumber-7.0.2/resources/PhoneNumberMetadata.xml#L940
and changed like this:

```
[...]
          4(?:
            [1235]|
            8[14-68]
          )|
[...]
```

dlyongemallo:
The document says that 3487 and 3489 are area codes. It also says that 3482 and 3483 are area codes. However, it also lists 348 as an area code. Is it ever possible to distinguish between the "348" area code which just happens to be followed by 2, 3, 7, or 9 and one of the "348[2379]" area codes, or does it never happen that "348" is followed by one of those digits when the area code is "348"? Can you point to the document and page where this is clarified, and also where in the specification it says 3487 and 3489 should be formatted as a block of four digits?

mdaloia:
I have found these documents:

1) http://www.cnc.gob.ar/numeracion-geografica_p137
It's in an excerpt in Spanish. In that page there is a link on the right called "National Numbering Plan" and send you to the PDF document of the point 2.

[...]

PR #618, opened 2015-02-18
Argentina - Invalid Format for Area Codes 3487 and 3489

mdaloia:
Implemented a fix for googlei18n/libphonenumber#611


hagbard:
This is now fixed by metadata update 7.0.3

PR #619, opened 2015-02-18
update libphonenumber issue links from Google Code to github

dlyongemallo:


dlyongemallo:
There's also a bunch of links to code.google.com in various release notes and config files, those should probably be updated as well.

PR #623, opened 2015-02-19
update code.google.com links to github in the maven pom.xml files, and in the demo jsp and html.

dlyongemallo:


dlyongemallo:
Per discussion with @jias0001, going to split this up into the maven change and other changes.

PR #626, opened 2015-02-19
Update debian/watch URL to point to new GitHub project.

roubert:
The old libphonenumber/libphonenumber has been abandoned in favour of
the new googlei18n/libphonenumber project.


roubert:
This URL in debian/ is not for human consumption and also never pointed to code.google.com but it points to a magic GitHub file that lists all tags and provides the mechanism for downloading source packages.

PR #629, opened 2015-02-20
update timezones/map_data.txt to more recent (2015-02-09) data

dlyongemallo:


resources/timezones/map_data.txt
```diff
@@ -1972,7 +1991,7 @@
 95|Asia/Rangoon
 960|Indian/Maldives
 961|Asia/Beirut
-962|Asia/Amman
+962|Asia/Riyadh
```
dlyongemallo: This seems wrong to me. 962 is Jordan, so I expected Amman rather than Riyadh here.

dlyongemallo:
It seems the +962 thing was an error in our data. I've fixed that, as well as updated to the 2015-02-29 data.

roes:
lgtm

Issue #555, opened 2014-12-03
Invalid as-you-type-formatter result for Samoa numbers


mihaelacr-google:
Imported from [Google Code issue #555](https://code.google.com/p/libphonenumber/issues/detail?id=555) created by [panteleymonov](https://code.google.com/u/108015422593122212925/) on 2014-11-13T16:08:10.000Z:

----
<b>What steps will reproduce the problem?</b>

try http://libphonenumber.googlecode.com/svn/trunk/javascript/i18n/phonenumbers/demo-compiled.html
Input number: +68535198

<b>What is the expected output? What do you see instead?</b>

Current output:
Result from isPossibleNumber(): true
International format: +685 35198
****AsYouTypeFormatter Results****
Char entered: + Output: +
Char entered: 6 Output: +6
Char entered: 8 Output: +68
Char entered: 5 Output: +685 
Char entered: 3 Output: +685 3
Char entered: 5 Output: +685 35
Char entered: 1 Output: +685351
Char entered: 9 Output: +6853519
Char entered: 8 Output: +68535198

--------------
Expected:

****AsYouTypeFormatter Results****
Char entered: + Output: +
Char entered: 6 Output: +6
Char entered: 8 Output: +68
Char entered: 5 Output: +685 
Char entered: 3 Output: +685 3
Char entered: 5 Output: +685 35
Char entered: 1 Output: +685 351
Char entered: 9 Output: +685 3519
Char entered: 8 Output: +685 35198

<b>What version of the product are you using? On what operating system?</b>

r715 libphonenumber v7.0, js version

<b>Please provide any additional information below.</b>

dlyongemallo:
The Argentine case appears to be unrelated. Please file a separate issue. Thanks.

keghani:
Samoa is fixed: http://libphonenumber.appspot.com/phonenumberparser?number=%2B68535198

For Argentina I filed https://github.com/googlei18n/libphonenumber/issues/1087

PR #631, opened 2015-02-23
In the as-you-type-formatter, insert a space after +685 (Samoa) when the number is a block. Issue #555

dlyongemallo:


dlyongemallo:
Committed as part of pull #637.

PR #634, opened 2015-02-24
Change name of testing class

evanstade:
from RegionCode to TestRegionCode. This avoids an ODR violation.

See http://crbug.com/456021


evanstade:
ping (roubert?)

keghani:
@roubert

keghani:
Summary:

RegionCode is defined in cpp both in [src/phonenumbers/region_code.h](https://github.com/googlei18n/libphonenumber/blob/master/cpp/src/phonenumbers/region_code.h) and [test/phonenumbers/test_util.h](https://github.com/googlei18n/libphonenumber/blob/master/cpp/test/phonenumbers/test_util.h).

This is preventing Chromium from enabling ODR violation everywhere.

@lararennie do you know of reasons why we want/ed these to be named the same?

lararennie:
One is actually used by the library. The other one is used for tests only. The right fix would be to delete the one in test utils and make the whole library use RegionCode (the other class) and make it contain all the regions we support.

roubert:
@keghani Sorry, I can't help you with any historical knowledge here, I've never been involved in this part of libphonenumber.

keghani:
Thanks for sending this @evanstade! Sorry it took us a while to triage all pull requests appropriately.

This should be fixed by us since we need to do some other changes first. I'm tracking at https://github.com/googlei18n/libphonenumber/issues/1852 and will close this.

PR #639, opened 2015-03-04
Fix validation for Belarusian operator

Stafox:
Here is a small fix which allow enter valid phone number from belarussian operator "Life:)" http://en.life.com.by/. Phone number from this operator have format: +375 25 XXX XX XX.


keghani:
Hi Stafox,

What's the problem you're trying to solve? According to http://www.eng.beltelecom.by/en/subscribers/phone-codes, the 25 prefix must be followed by 5, 6, 7, or 9. This is already supported.

If you have a number outside these ranges that you believe should be supported, please raise a bug with evidence.

When contributing a patch for a new/updated range, you should only modify one file: resources/PhoneNumberMetadata.xml. Other files like javascript/i18n/phonenumbers/metadata.js will get generated from this during release. The .xml's format also makes it easier to see what's supported. In this case the relevant metadata is:
      <mobile>
        <nationalNumberPattern>
          (?:
            2(?:
              5[5679]|
              9[1-9]
            )|
            33\d|
            44\d
          )\d{6}
        </nationalNumberPattern>
        <possibleNumberPattern>\d{9}</possibleNumberPattern>
        
        <exampleNumber>294911911</exampleNumber>
      </mobile>

Closing this. Please reopen if I can help further.

keghani:
I meant:


      <mobile>
        <nationalNumberPattern>
          (?:
            2(?:
              5[5679]|
              9[1-9]
            )|
            33\d|
            44\d
          )\d{6}
        </nationalNumberPattern>
        <possibleNumberPattern>\d{9}</possibleNumberPattern>
        
        <exampleNumber>294911911</exampleNumber>
      </mobile>

Stafox:
Hello,

Yes here is my mistake. I got unvalidated infromation from customer about case when phone number with prefix +375 25 may starts from 2. So, you right, according to beltelecom/phone-codes. Thanks for your time :)

keghani:
You're welcome, thanks for getting back to us :)

PR #640, opened 2015-03-04
Updated README.md for latest version and removed broken line to googlecode for JS version.

hagbard:
We cannot host the compiled demo on github as github will not 'serve' html pages directly. This needs to be moved to somewhere like appspot and updated with the Java demo.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,10 @@ Google's common Java, C++ and Javascript library for parsing, formatting, storin
   * ` PhoneNumberOfflineGeocoder ` - provides geographical information related to a phone number.
   * ` PhoneNumberToCarrierMapper ` - provides carrier information related to a phone number.
 
-# Demo
-[Java Version](http://libphonenumber.appspot.com/) (running code r7.0.3 (on Google Code))
+# Demo (v7.0.3)
+[Java](http://libphonenumber.appspot.com/)
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -15,9 +15,7 @@ Google's common Java, C++ and Javascript library for parsing, formatting, storin
   * ` PhoneNumberToCarrierMapper ` - provides carrier information related to a phone number.
 
 # Demo
-[Java Version](http://libphonenumber.appspot.com/) (running code r729 (on Google Code))
-
-[JavaScript Version](http://libphonenumber.googlecode.com/svn/trunk/javascript/i18n/phonenumbers/demo-compiled.html) (running JS code r729 (on Google Code))
```
jias0001: This can be done. Simply searching on Google for "open html file from github" reveals the solution, which is to replace this with:

https://rawgit.com/googlei18n/libphonenumber/master/javascript/i18n/phonenumbers/demo-compiled.html

In case you want more details: http://rawgit.com/

hagbard: Thanks. I won't do this now as the build is borken anyway, so this will be submitted when it's all fixed.

jias0001: Sure, let me know when you want me to take another look later.

hagbard:
Updated, simplified a bit and check that they work.

Issue #658, opened 2015-03-30
asyoutypeformatter_test.js cannot be run from github cleanly.


hagbard:
Running the as-you-type formatter tests is currently broken. On a clean clone of the repository you get.

     [exec] i18n/phonenumbers/asyoutypeformatter_test.js:1164: ERROR - Access to private property getExtractedNationalPrefix_ of i18n.phonenumbers.AsYouTypeFormatter not allowed here.
     [exec]   assertEquals('0', f.getExtractedNationalPrefix_());
     [exec]                     ^
     [exec] 
     [exec] i18n/phonenumbers/asyoutypeformatter_test.js:1166: ERROR - Access to private property getExtractedNationalPrefix_ of i18n.phonenumbers.AsYouTypeFormatter not allowed here.
     [exec]   assertEquals('', f.getExtractedNationalPrefix_());
     [exec]                    ^
     [exec] 
     [exec] 2 error(s), 0 warning(s), 97.6% typed
     [exec] calcdeps.py: JavaScript compilation failed.

This seems related to how the JS compiler interprets private functions.

dlyongemallo:
@hagbard - Can this be closed?

PR #659, opened 2015-03-30
Fixing tests

hagbard:
Workaround/fix for: https://github.com/googlei18n/libphonenumber/issues/658

### DIFF
--- a/javascript/i18n/phonenumbers/asyoutypeformatter_test.js
+++ b/javascript/i18n/phonenumbers/asyoutypeformatter_test.js
@@ -1150,6 +1150,7 @@ function testAYTFShortNumberFormattingFix_US() {
   assertEquals('1 22', f.inputDigit('2'));
 }
 
+// https://github.com/googlei18n/libphonenumber/issues/658
 // TODO: Re-enable this test when it can be run from a git repository without:
 // [exec] i18n/phonenumbers/asyoutypeformatter_test.js:1164:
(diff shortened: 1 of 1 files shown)


javascript/i18n/phonenumbers/asyoutypeformatter_test.js
```diff
@@ -1150,37 +1150,44 @@ function testAYTFShortNumberFormattingFix_US() {
...
-  var f = new i18n.phonenumbers.AsYouTypeFormatter(RegionCode.KR);
-
-  // Check that when we have successfully extracted an IDD, the previously
-  // extracted NDD is cleared since it is no longer valid.
-  assertEquals('0', f.inputDigit('0'));
-  assertEquals('00', f.inputDigit('0'));
-  assertEquals('007', f.inputDigit('7'));
-  assertEquals('0070', f.inputDigit('0'));
-  assertEquals('00700', f.inputDigit('0'));
-  assertEquals('0', f.getExtractedNationalPrefix_());
```
jias0001: You just need to comment out this line and line 1166, which are causing issues. We still want this test.

jias0001: Or better, just remove these two lines. I don't see how this can work with closure strict compilation mode.

jias0001:
It is best we just remove that two lines. In that case we don't need the comment anymore, and can mark the bug as fixed.

PR #660, opened 2015-04-02
Small metadata update for AS, IN & NE

alecx:
American Samoa - Blue sky is now using 1-684-272 -
http://localcallingguide.com/lca_prefix.php?npa=684
examples:
168427216xx
168427252xx

India - 7210 is being used by Aircel
examples:
9172104311xx
9172104313xx
9172104348xx

Niger - Airtel introduced 88 prefix
examples:
227881204xx
227881263xx
227881333xx
227881340xx


dlyongemallo:
Thanks for the report! We've updated American Samoa and Niger, so these should be fixed in the next push. Still investigating India.

keghani:
Fixed by release 7.0.11.

PR #664, opened 2015-04-13
Fix English names of Finnish numbering areas Uusimaa and Häme

anssih:
Fix the English names of Finnish numbering areas Uusimaa and Häme
in geocoding list according to the official numbering plan:
http://www.viestintavirasto.fi/attachments/maaraykset/Regulation_32.pdf

The previous names were seemingly user-made translations that are not
actually in use (confirmed by Google search as well), but were listed in
Wikipedia (with no references/sources).

The changes are:
Tavastia => Häme
Nylandia => Uusimaa

I've made this PR on top of PR #566 since otherwise they would conflict.


anssih:
Rebased on top of updated 566.

Also, updated the commit message. I found some actual uses of "Häme = Tavastia" on internet after all, though still nothing official.

I still think Häme would be better, since it is used by all official sources for the region, e.g. the Regional Councils of Häme in English:
http://www.hameenliitto.fi/en/hame-region

keghani:
Hi, sorry this took so long...

If you can merge the changes from upstream I'll approve and merge this as-is.
Since you signed the CLA, if you can't get to it just let us know and we'll implement it.

Thanks for your patience!

anssih:
Merged changes from upstream as requested, no conflicts. Let me know if want me to rebase instead.

keghani:
Nope, this is great, thanks!

PR #674, opened 2015-04-21
9012345678 is a better mobile example for Japan

johnnyshields:
The current JP mobile example number 7012345678, while technically valid, looks strange to most Japanese people. 070- mobile numbers are reserved for PHS (Personal Handy-phone System) which is a legacy technology that is not commonly used. 090- are more common.

### DIFF
--- a/resources/PhoneNumberMetadata.xml
+++ b/resources/PhoneNumberMetadata.xml
@@ -12760,7 +12760,7 @@
       <mobile>
         <nationalNumberPattern>[7-9]0[1-9]\d{7}</nationalNumberPattern>
         <possibleNumberPattern>\d{10}</possibleNumberPattern>
-        <exampleNumber>8012345678</exampleNumber>
+        <exampleNumber>9012345678</exampleNumber>
       </mobile>
       <pager>
         <nationalNumberPattern>20\d{8}</nationalNumberPattern>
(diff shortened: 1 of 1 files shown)


dlyongemallo:
In the ITU doc, there is no difference between 70, 80, and 90:
http://www.itu.int/dms_pub/itu-t/oth/02/02/T020200006D0001PDFE.pdf

On the Internet, the three prefixes seem to be used with about the same order of magnitude in frequency:
https://www.google.com/webhp#q=%22%2B81+70+1000..9999+1000..9999%22+site:.jp
https://www.google.com/webhp#q=%22%2B81+80+1000..9999+1000..9999%22+site:.jp
https://www.google.com/webhp#q=%22%2B81+90+1000..9999+1000..9999%22+site:.jp

It seems many Japanese web sites are listing 70 numbers. Do you have any specific evidence that this "looks strange to most Japanese people"?

johnnyshields:
I have lived in Japan for 8 years and am CTO of a company which offers smartphone apps and phone-based voice response services.

So, the first bit of evidence is what you posted that 090 is clearly the most common of the three.

070 are reserved for PHS-style phones, which are the type of phone used by Gordon Gekko:
image

If you give a girl your number and its 070, she will probably laugh at you, unless you're also into vinyl and retro cameras in which case it will prove your hipster bonafides.

https://en.wikipedia.org/wiki/Personal_Handy-phone_System
https://www.ntt-east.co.jp/en/product/other_phs.html

dlyongemallo:
Ok, I'll make the change. :-)  It'll be in the next release.

johnnyshields:
Great thank you!

andyst:
Thanks! This is in libphonenumber 7.0.7.

PR #681, opened 2015-04-28
dbeaumont update 7.0.5

hagbard:
Updates from the Maven push from last week's update.
This includes one extra change for Indonesia which got picked up by the script (unexpectedly).
We include this here to have consistency with the JARs pushed to Maven.


jias0001:
LGTM.

PR #690, opened 2015-05-15
Throw errors instead of strings

ruimarinho:
Modified the javascript port to throw errors instead of strings.

### DIFF
--- a/javascript/i18n/phonenumbers/phonenumberutil.js
+++ b/javascript/i18n/phonenumbers/phonenumberutil.js
@@ -2953,7 +2953,7 @@ i18n.phonenumbers.PhoneNumberUtil.prototype.getCountryCodeForRegion =
  *     calling code for.
  * @return {number} the country calling code for the region denoted by
  *     regionCode.
- * @throws {string} if the region is invalid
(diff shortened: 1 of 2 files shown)


jias0001:
Sorry, cannot accept this, as it changes the public interface and can break clients' code which relies on the method throwing strings.

ruimarinho:
@jias0001 would it be possible to accept this merge when releasing v8.0.0?

jias0001:
We don't want to change the public API unless there is a very good reason behind it, because it requires all clients to update their code. This change doesn't seem to be something that is worth the trouble of all clients updating their code.

ruimarinho:
@jias0001 on PR https://github.com/googlei18n/libphonenumber/pull/595 you mentioned *I remember sometime last year we have updated our code to mostly throwing Error instead of string. Please feel free to send us a patch to update the rest if they are missed from that update.*.

Since this change won't be accepted due to BC concerns, then maybe the npm module being listed could be re-considered? It can become quite frustrating to debug errors thrown as strings.

ruimarinho:
Superseded by https://github.com/googlei18n/libphonenumber/pull/1459.

PR #693, opened 2015-05-18
Telenor MM new range

alecx:
Telenor Myanmar introduced a new number series in market starting with
+95978.


andyst:
Thanks for the report! Added to our internal bug triage queue. There are currently a lot of updates (also from ITU), so it might take a bit longer.

andyst:
Was a bit fast there...

Do you have any evidence (government or from the telco) for this change?

roes:
This will be part of the next release

andyst:
This change is in libphonenumber 7.0.6 which we released a few weeks ago.

PR #694, opened 2015-05-18
Telenor PK new range

alecx:
Telenor PK introduced a new number series +92-348


andyst:
Hi, do you have any government / telco docs as evidence for the change?

Thanks!

roes:
This will be part of the next release

andyst:
This change is in libphonenumber 7.0.6 which we released a few weeks ago.

PR #716, opened 2015-06-09
Update Readme to include node.js

harshil93:


andyst:
Thanks. Based on a previous response / policy, we can't make this change though.

See:
https://github.com/googlei18n/libphonenumber/pull/595#issuecomment-77531156

PR #718, opened 2015-06-17
Fix NullPointerException when ShortNumberInfo.isPossibleShortNumber is used with a region not in ShortNumberMetadata

giggsey:
Example code:

```java
import com.google.i18n.phonenumbers.*;
import com.google.i18n.phonenumbers.Phonenumber.PhoneNumber;

public class Test {

	/**
	 * @param args
	 */
	public static void main(String[] args) {
        PhoneNumber phoneNumber = null;
        PhoneNumberUtil phoneNumberUtil = PhoneNumberUtil.getInstance();
        phoneNumber = phoneNumberUtil.getExampleNumber("NE");

        ShortNumberInfo shortNumberInfo = ShortNumberInfo.getInstance();

        shortNumberInfo.isPossibleShortNumber(phoneNumber);

	}

}
```

`NE` is a valid region, and exists in PhoneNumberMetadata. This means that the check in [`getRegionCodesForCountryCode`](https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/ShortNumberInfo.java#L97) returns the correct region code.

It then loops through, loading the metadata for each region, but doesn't actually check whether the metadata exists.

Credit to @RuslanGetmansky via https://github.com/giggsey/libphonenumber-for-php/issues/68

I haven't updated any unit tests for this, but I can do if needed (my Java is not very good)

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/ShortNumberInfo.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/ShortNumberInfo.java
@@ -156,6 +156,9 @@ public class ShortNumberInfo {
     String shortNumber = getNationalSignificantNumber(number);
     for (String region : regionCodes) {
       PhoneMetadata phoneMetadata = MetadataManager.getShortNumberMetadataForRegion(region);
+      if (phoneMetadata == null) {
+        continue;
+      }
       if (matcherApi.matchesPossibleNumber(shortNumber, phoneMetadata.getGeneralDesc())) {
         return true;
       }


dlyongemallo:
Thanks for reporting this, it'll be fixed in the next release.

andyst:
Thanks! This is in libphonenumber 7.0.7.

giggsey:
Is it?

[ShortnumberInfo.java](https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/ShortNumberInfo.java) hasn't changed since Dec 2014.

andyst:
Thanks! Looks like we didn't make the change on github yet. Reopening.

giggsey:
I can see this has been fixed by #731, closing this one.

PR #741, opened 2015-07-10
Fix Wunused-private-field warning in debug builds.

nico:


pliard-chromium:
LGTM, thanks! I don't seem to have write access. @dlyongemallo can you merge this in?

PR #760, opened 2015-07-24
Changes the format of Brazilian nine digits mobile numbers...

aricaldeira:
… to the more readable 999-999-999 instead of 99999-9999

Since July 2012 Brazilian mobile numbers have been changing from the old format:
AA99999999 : 2 digits for the area code, 8 digits for the actual number
to
AA**9**99999999: 2 digits for the area code, a fixed (for now) digit **9**, and the old 8 digits for the number

The national format in this library stated the those 9 digits numbers should be formated as:
(AA) **9**9999-9999

That format makes most numbers quite unreadable, since Brazilian mobile numbers, in most States, already started with a 9, making several numers like:
(11) 99995-3214 (invented number)

The proposal changes the format to the much more readable:
(11) 999-953-214
libphonenumber-brazilian-mobile-numbers

### DIFF
--- a/resources/PhoneNumberMetadata.xml
+++ b/resources/PhoneNumberMetadata.xml
@@ -3163,14 +3163,14 @@
           <format>$1-$2</format>
           <intlFormat>NA</intlFormat>
         </numberFormat>
-        <numberFormat nationalPrefixFormattingRule="$FG" pattern="(\d{5})(\d{4})">
+        <numberFormat nationalPrefixFormattingRule="$FG" pattern="(\d{3})(\d{3})(\d{3})">
           <leadingDigits>
             9(?:
               [1-9]|
               0[1-9]
             )
           </leadingDigits>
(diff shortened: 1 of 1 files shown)


bmarry:
Hi 

I see this is a neater format for readability but do you have some examples or official sites that format numbers in this way in Brazil please, for example government departments, hospitals, large stores etc please to ensure this is a standard format

Thanks 
Brendan

dlyongemallo:
The ITU has the format XXXXX-XXXX:
http://www.itu.int/dms_pub/itu-t/oth/02/02/T020200001D0001PDFE.pdf

It is also the more common format online. Compare:
https://www.google.ch/#q=%22%2B55+900..999+100..999%22
https://www.google.ch/#q=%22%2B55+90000..99999+1000..9999%22

I would suggest making the recommendation to the ITU. As of now, the evidence is insufficient for us to change the format.

PR #776, opened 2015-08-10
Kenya mobiles are always 9 digit national numbers

lincolnq:
Kenya mobile phone numbers are always exactly 9 digits after the country code. (Wikipedia used to think that they could be 10 digits, but that is wrong; I fixed that also.)

https://en.wikipedia.org/wiki/Telephone_numbers_in_Kenya#Mobile_phone_numbers

At this time the Communication Authority of Kenya's website (http://www.ca.go.ke/) is offline but I contacted Safaricom (a major mobile operator in Kenya) and they confirmed this on their support line. I also confirmed it with two friends who live & work in Kenya.

(Apologies if this is the wrong way to submit patches; I wasn't able to figure out the Rietveld code review tool.)

### DIFF
--- a/resources/PhoneNumberMetadata.xml
+++ b/resources/PhoneNumberMetadata.xml
@@ -12962,7 +12962,7 @@
           <leadingDigits>[24-6]</leadingDigits>
           <format>$1 $2</format>
         </numberFormat>
-        <numberFormat pattern="(\d{3})(\d{6,7})">
+        <numberFormat pattern="(\d{3})(\d{6})">
           <leadingDigits>7</leadingDigits>
           <format>$1 $2</format>
         </numberFormat>


andyst:
The latest ITU doc supports what you're saying, but it's from 2010, i.e. a bit old.
http://www.itu.int/dms_pub/itu-t/oth/02/02/T02020000700001PDFE.pdf

http://www.ca.go.ke/ works for me right now. Is there a page / download on that website which provides supportive evidence for the requested change?

Thanks!

andyst:
@penmetsaa Can you confirm that this request (all numbers should be 9 digits) based on the 2013 national number plan PDF we have?

penmetsaa:
Hi Andy,

As per ITU document(2013 version) all mobile prefixes have length exactly 9 digits.
And it is implemented in same way in our metadata as well.
For double confirmation I have tested with 10digit mobile number, which is resulted as invalid number. 

Thanks & Regards,
Aravind

keghani:
Fixed in release 7.0.11.

PR #790, opened 2015-08-14
Fix phone number match.

renatofilho:
Trying to match: "+28909" with "+13611361530" should return  "NO_MATCH";


andyst:
Thanks for the proposal. I think the fix needs to be different. The pre-existing code comment right above this, at line 2733, seems wrong to me. INVALID_COUNTRY_CODE_ERROR doesn't mean "The second number has no country calling code.", we also get INVALID_COUNTRY_CODE_ERROR when the number has a country code but it's invalid, as it is in the case of "+28909".

Thus looking at second_number_with_first_number_region isn't the right spot to fix this IMO. On top of handling second_number_with_first_number_region as it is done right now, we also need to have some code to handle the case of "we got INVALID_COUNTRY_CODE_ERROR because the number has a country code but it's not valid."

keghani:
I'll close this as the fix needs to be different and we're tracking the issue at #791.

PR #794, opened 2015-08-19
replaced Novafone prefix 5 to 555 by Liberia

spleenjack:
> NOVAFONE, in accordance with the directives of the Liberia Telecommunications Authority(LTA) launched on Sept 19th the new ten digits numbering plan.

> According to a communication from the company the previous "05" will be replaced by "0555". Novafone said it is very important to note that subscribers need to add "55" to the existing code of "05", followed by the NOVAFONE number.

The sources are:
http://allafrica.com/stories/201309232245.html
http://allafrica.com/stories/201309232166.html

This update has been reflected already on the ITU document about numbering plan of Liberia: http://www.itu.int/oth/T0202000079/en

### DIFF
--- a/resources/PhoneNumberMetadata.xml
+++ b/resources/PhoneNumberMetadata.xml
@@ -14489,7 +14489,8 @@
         <nationalNumberPattern>
           2\d{7}|
           [37-9]\d{8}|
-          [45]\d{6}
+          4\d{6}|
+          5\d{8}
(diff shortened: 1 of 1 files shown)


roes:
Thanks for letting us know, we'll look into it.

keghani:
Fixed in release 7.0.11.

PR #813, opened 2015-09-10
Fix: javascript README formatting by adding md extension

alucic:
This PR just adds extension to readme file under javascript folder and fixes markdown formatting


andyst:
Thanks, good idea.

The diff makes it look like we're adding / deleting unrelated files. Can you use `git mv README README.md` if you haven't done so already?

See:
http://githowto.com/moving_files

The goal is to preserve the history of the file and to also make the code review here easier by having a nice visual diff of the changes between the two versions of the file.

alucic:
@andyst 

Unfortunately using `git mv README README.md` or  `mv README README.md` `git rm README` `git add README.md` results in the same action, new file is being created, history is not preserved

I'm not sure if there's a way around it

mihaelacr-google:
Changes look good to me. Thanks for your contribution!

PR #823, opened 2015-09-18
Added israel as a region

weitzhandler:


dlyongemallo:
Why do you want to add Israel to the test data? Is there something special about Israel that you want to test, which isn't already covered by the logic for other countries? If so, please actually add a test.

keghani:
Hello!

The file you have changed is actually a test file. It only includes some regions that we're testing.

The implementation does contain Israel: https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/CountryCodeToRegionCodeMap.java

If you meant to add a test (see @dlyongemallo's comment), please re-open.

weitzhandler:
Sorry alright I get it now.

PR #828, opened 2015-09-22
Old JDK support

dpocock:
These are some patches we had in the Debian version to support older JDKs.  I realize older JDKs are deprecated but it is relatively easy to support them with changes like this.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/828)

### DIFF
--- a/java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
+++ b/java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
@@ -200,7 +200,7 @@ public class PhoneNumberParserServlet extends HttpServlet {
       } else {
         appendLine("Result from isValidNumber()", Boolean.toString(isNumberValid), output);
         if (isNumberValid) {
-          if (!defaultCountry.isEmpty() && defaultCountry != "ZZ") {
+          if (!defaultCountry.length() == 0 && defaultCountry != "ZZ") {
             appendLine(
                 "Result from isValidNumberForRegion()",
                 Boolean.toString(phoneUtil.isValidNumberForRegion(number, defaultCountry)),
--- a/java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java
+++ b/java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java
@@ -172,7 +172,7 @@ public class PhoneNumberToTimeZonesMapper {
    */
   private List<String> getTimeZonesForGeocodableNumber(PhoneNumber number) {
     List<String> timezones = prefixTimeZonesMap.lookupTimeZonesForNumber(number);
-    return Collections.unmodifiableList(timezones.isEmpty() ? UNKNOWN_TIME_ZONE_LIST
+    return Collections.unmodifiableList(timezones.size() == 0 ? UNKNOWN_TIME_ZONE_LIST
                                                             : timezones);
   }
 
@@ -185,7 +185,7 @@ public class PhoneNumberToTimeZonesMapper {
    */
   private List<String> getCountryLevelTimeZonesforNumber(PhoneNumber number) {
     List<String> timezones = prefixTimeZonesMap.lookupCountryLevelTimeZonesForNumber(number);
-    return Collections.unmodifiableList(timezones.isEmpty() ? UNKNOWN_TIME_ZONE_LIST
+    return Collections.unmodifiableList(timezones.size() == 0 ? UNKNOWN_TIME_ZONE_LIST
                                                             : timezones);
   }
 }
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java
@@ -110,7 +110,7 @@ final class MultiFileMetadataSourceImpl implements MetadataSource {
       in = new ObjectInputStream(source);
       PhoneMetadataCollection metadataCollection = loadMetadataAndCloseInput(in);
       List<PhoneMetadata> metadataList = metadataCollection.getMetadataList();
-      if (metadataList.isEmpty()) {
+      if (metadataList.size() == 0) {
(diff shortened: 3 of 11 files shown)


java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java
```diff
@@ -172,7 +172,7 @@ public static String getUnknownTimeZone() {
    */
   private List<String> getTimeZonesForGeocodableNumber(PhoneNumber number) {
     List<String> timezones = prefixTimeZonesMap.lookupTimeZonesForNumber(number);
-    return Collections.unmodifiableList(timezones.isEmpty() ? UNKNOWN_TIME_ZONE_LIST
+    return Collections.unmodifiableList(timezones.size() == 0 ? UNKNOWN_TIME_ZONE_LIST
```
lararennie: This is a List, which was introduced in Java 1.2, at which time isEmpty was already there?
the isEmpty on string was indeed introduced in Java 1.6, so that we should avoid.

Was this an accidental "fix"? I notice PhoneNumberUtil uses isEmpty on lists, and that isn't listed here?

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -200,7 +200,7 @@ private StringBuilder getOutputForSingleNumber(
       } else {
         appendLine("Result from isValidNumber()", Boolean.toString(isNumberValid), output);
         if (isNumberValid) {
-          if (!defaultCountry.isEmpty() && defaultCountry != "ZZ") {
```
lararennie: This is the demo, which requires appengine, and I don't think you can run that without using java 7 anyway?

java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java
```diff
@@ -110,7 +110,7 @@ void loadMetadataFromFile(String filePrefix, String regionCode, int countryCalli
       in = new ObjectInputStream(source);
       PhoneMetadataCollection metadataCollection = loadMetadataAndCloseInput(in);
       List<PhoneMetadata> metadataList = metadataCollection.getMetadataList();
-      if (metadataList.isEmpty()) {
+      if (metadataList.size() == 0) {
```
lararennie: same as previous comment

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberUtilTest.java
```diff
@@ -1220,7 +1220,7 @@ public void testGetRegionCodesForCountryCode() {
     assertTrue(phoneUtil.getRegionCodesForCountryCode(49).contains(RegionCode.DE));
     assertTrue(phoneUtil.getRegionCodesForCountryCode(800).contains(RegionCode.UN001));
     // Test with invalid country calling code.
-    assertTrue(phoneUtil.getRegionCodesForCountryCode(-1).isEmpty());
+    assertTrue(phoneUtil.getRegionCodesForCountryCode(-1).size() == 0);
```
lararennie: and here

tools/java/common/src/com/google/i18n/phonenumbers/BuildMetadataFromXml.java
```diff
@@ -137,7 +137,7 @@ public static PhoneMetadataCollection buildPhoneMetadataCollection(String inputX
       } else {
         // For most countries, there will be only one region code for the country calling code.
         List<String> listWithRegionCode = new ArrayList<String>(1);
-        if (!regionCode.isEmpty()) {  // For alternate formats, there are no region codes at all.
+        if (!regionCode.length() == 0) {  // For alternate formats, there are no region codes at all.
```
lararennie: We changed this already at head to compare to empty string.

tools/java/common/src/com/google/i18n/phonenumbers/BuildMetadataFromXml.java
```diff
@@ -202,7 +202,7 @@ static String getNationalPrefix(Element element) {
             validateRE(element.getAttribute(NATIONAL_PREFIX_TRANSFORM_RULE)));
       }
     }
-    if (!nationalPrefix.isEmpty()) {
+    if (!nationalPrefix.length() == 0) {
```
lararennie: For consistency, we could change this to check it's not the empty string, like we did for the region code comparison. Or otherwise, change it there to the length check.

tools/java/cpp-build/test/com/google/i18n/phonenumbers/BuildMetadataCppFromXmlTest.java
```diff
@@ -175,11 +176,11 @@ public MockedCommand(String expectedInputFilePath, boolean expectedLiteMetadata,
       assertEquals(expectedVariant, variant);
       return sourceOut;
     }
-    String capturedHeaderFile() {
-      return new String(headerOut.toByteArray(), UTF_8);
+    String capturedHeaderFile() throws UnsupportedEncodingException {
+      return new String(headerOut.toByteArray(), "UTF-8");
```
lararennie: Okay with the change, but in that case, you need to remove line 144 since it's not used anymore.

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -196,7 +196,7 @@ private static void writeCountryCallingCodeMappingToJavaFile(
     // calling codes listed in it.
     boolean hasRegionCodes = false;
     for (List<String> listWithRegionCode : countryCodeToRegionCodeMap.values()) {
-      if (!listWithRegionCode.isEmpty()) {
+      if (!listWithRegionCode.size() == 0) {
```
lararennie: same qn here and below about lists

keghani:
I'm closing as this issue is infeasible for us to fix at this point. Please re-open if you respond so we pick up on it.

PR #831, opened 2015-09-24
C++ symbols map

dpocock:


keghani:
Thanks @dpocock!

Fixing this issue, #828, and #830 may require us changing other systems that depend on this library, which we can't do overnight. It's not our top priority to look into this at the moment. Please bear with us and feel free to contribute a patch to your own fork. I'll keep the PRs open to revisit when we have some free cycles.

keghani:
Hi and sorry for the late turnaround,

Would you mind filing and issue describing what the problem is that this PR attempts to fix?

keghani:
I'm closing as this issue is infeasible for us to fix at this point. Please re-open if you respond so we pick up on it.

PR #833, opened 2015-09-25
Implementation of SingleFileMetadata reader.

mirunabarbu:
Implementation of SingleFileMetadata reading with files created exclded from jar.


keghani:
Please remove ```tools/java/cpp-build/target/cpp-build-1.0-SNAPSHOT-jar-with-dependencies.jar```, the contents don't change as a result of adding ```tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java```. Sorry I hadn't seen this before.

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -133,7 +140,7 @@ public boolean start() {
     }
 
     if (inputFile == null ||
-        outputDir == null ||
+        ((outputDir == null) == (outputFile == null)) ||
```
dlyongemallo: Is the interior "==" supposed to be "||"? Or "&&"?

If it's indeed meant to be "==" (i.e., they are both null or both non-null), there should be a comment to that effect, or the logic rewritten to make that clear.

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -141,24 +148,33 @@ public boolean start() {
...
+          out.close();
         }
-        PhoneMetadataCollection outMetadataCollection = new PhoneMetadataCollection();
-        outMetadataCollection.addMetadata(metadata);
-        FileOutputStream outputForRegion = new FileOutputStream(filePrefix + "_" + regionCode);
-        ObjectOutputStream out = new ObjectOutputStream(outputForRegion);
-        outMetadataCollection.writeExternal(out);
+      }
+
+      if (outputFile != null) {
```
dlyongemallo: By the above, I think this cannot be non-null at this point?

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
dlyongemallo: Discussed in person: this is supposed to check that one and only one of them is non-null.

java/libphonenumber/src/com/google/i18n/phonenumbers/SingleFileMetadataSourceImpl.java
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * Loads the metadata protocol buffer from the given stream and closes the stream afterwards. Any
+   * exceptions that occur while reading the stream are propagated (though exceptions that occur
+   * when the stream is closed will be ignored).
+   *
+   * @param source  the non-null stream from which metadata is to be read.
+   * @return        the loaded metadata protocol buffer.
+   */
+  private static PhoneMetadataCollection loadMetadataAndCloseInput(InputStream source) {
```
dlyongemallo: Slightly inconsistent with MultiFileMetadataSourceImpl.java here: there, loadMetadataAndCloseInput takes in an ObjectInputStream already. It doesn't matter too much since these are private, but it'd be nice if they were consistent.

keghani:
FTR I will reiterate on this to make SingleFileMetadataSourceImpl.java consistent with MultiFileMetadataSourceImpl.java. It's blocked on some other changes which is why I asked Miruna to go ahead with this.

keghani:
Also FWIW the contents of tools/java/cpp-build/target/cpp-build-1.0-SNAPSHOT-jar-with-dependencies.jar did not change in this commit/PR.

keghani:
Also java/libphonenumber/src/com/google/i18n/phonenumbers/data/PhoneNumberMetadataProtoFile needed to be updated since the metadata changed between generating it for this commit and merging the PR - I did this separately at https://github.com/googlei18n/libphonenumber/commit/eb82ee3557fe53b8211f39189438b23fde875dc8

PR #835, opened 2015-09-28
Fix generation of java/CountryCodeToRegionCodeMap.java…

keghani:
… as side-effect to single file metadata generation.


tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -61,16 +61,16 @@
       "\n" +
       "  --" + INPUT_FILE + "=PATH     Read phone number metadata in XML format from PATH.\n" +
       "  --" + OUTPUT_DIR + "=PATH     Use PATH as the root directory for output files.\n" +
-      "  --" + OUTPUT_FILE + "=PATH    Writes to PATH.\n" +
-      "  Only 1 of " + OUTPUT_DIR + " or " + OUTPUT_FILE + " must be specified.\n" +
       "  --" + DATA_PREFIX +
           "=PATH    Use PATH (relative to " + OUTPUT_DIR + ") as the basename when\n" +
-      "                        writing phone number metadata (one file per region) in\n" +
-      "                        proto format.\n" +
+      "                        writing phone number metadata in proto format.\n" +
```
dlyongemallo: Clarify here that one file per region will be written into the directory with that prefix if SINGLE_FILE is false, otherwise one file named as just this prefix will be written.

PR #836, opened 2015-09-28
Rename PhoneNumberMetadataProtoFile...

keghani:
... to SingleFilePhoneNumberMetadataProto to avoid prefix conflict with per-region files (see https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java that uses PhoneNumberMetadataProto as the prefix).
Remove PhoneNumberMetadataProtoForTestingFile as it's not used anywhere.
Fix usage message at BuildMetadataProtoFromXml.java.


keghani:
I'll send another PR - I'll keep the single file for testing.

PR #839, opened 2015-09-29
Rename single file metadata to avoid conflict...

keghani:
... with per-region files' prefix; see:
- https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java
- https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/test/com/google/i18n/phonenumbers/TestMetadataTestCase.java


dlyongemallo:
What's conflicting with what?

java/build.xml
```diff
@@ -45,14 +45,14 @@
     </exec>
   </target>
 
-  <target name="build-phone-metadata-to-one-file">
+  <target name="build-single-file-phone-metadata">
```
dlyongemallo: Would it make sense to rename the other rule too, e.g., to "build-multi-file-phone-metadata"? And then have a comment above them both that exactly one of them should be used? (Same comment re: test targets.)

keghani:
PhoneNumberMetadataProtoFile, the single file with all regions' phone number metadata, has "PhoneNumberMetadataProto" as a prefix, which is what MultiFileMetadataSourceImpl looks for.
So I renamed PhoneNumberMetadataProtoFile to SingleFilePhoneNumberMetadataProto.

PhoneNumberMetadataProtoForTestingFile, the single file with all regions' test metadata, has "PhoneNumberMetadataProtoForTesting" as a prefix, which is what TestMetadataTestCase looks for. So I renamed PhoneNumberMetadataProtoForTestingFile to SingleFilePhoneNumberMetadataProtoForTesting.

The test metadata is currently not in use, I'm planning an upcoming change to do that.

java/build.xml
keghani: I don't know whether they're in use.

dlyongemallo: Okay, let's leave it then.

PR #870, opened 2015-10-29
merge from master into nanoproto (code only)

dlyongemallo:


keghani:
Please regenerate checked-in jar files (java-build and cpp-build; anything else?) before making more changes.

dlyongemallo:
As discussed, we won't be checking in generated files until ready to merge back into master (though we'll check them locally).

PR #880, opened 2015-11-02
Use protoc from java/lib rather than Maven repository.

dlyongemallo:


java/lib/com/google/protobuf/protobuf-java/maven-metadata-local.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<metadata>
+  <groupId>com.google.protobuf</groupId>
+  <artifactId>protobuf-java</artifactId>
+  <versioning>
+    <release>2.6.1</release>
+    <versions>
+      <version>2.6.1</version>
+    </versions>
+    <lastUpdated>20151102180249</lastUpdated>
```
keghani: What time zone is this supposed to be? UTC? I wonder why it's not in epoch seconds.

java/libphonenumber/pom.xml
```diff
@@ -46,4 +46,18 @@
     </plugins>
   </build>
 
+  <repositories>
+    <repository>
+      <id>protobuf-java-lib</id>
+      <url>file://${project.basedir}/../lib</url>
+    </repository>
+  </repositories>
+  <dependencies>
```
keghani: Thanks for moving this out of java/pom.xml :)

tools/java/common/pom.xml
```diff
@@ -49,6 +49,12 @@
...
 
+  <repositories>
+    <repository>
+      <id>protobuf-java-lib</id>
+      <url>file://${project.basedir}/../../../java/lib</url>
+    </repository>
+  </repositories>
   <dependencies>
     <dependency>
       <groupId>junit</groupId>
```
keghani: Fun fact: we have junit-4.8.1.jar under java/lib. I wonder what Maven is using, and who is using the one under java/lib.

keghani:
Merging, but I still want to discuss above points.

java/lib/com/google/protobuf/protobuf-java/maven-metadata-local.xml
dlyongemallo: As discussed, this is used by maven to check against, e.g., ~/.m2/repository/com/google/protobuf/protobuf-java/2.6.1/protobuf-java-2.6.1.jar.lastUpdated - to determine whether a local version is newer or older than the one checked into our repo.

tools/java/common/pom.xml
dlyongemallo: There are some path issues with the junit that's checked in. We should fix that.

PR #884, opened 2015-11-05
remove check for nationalPrefixFormattingRule empty before setting nationalPrefixOptionalWhenFormatting in BuildMetadataFromXml

dlyongemallo:
Verified results in no jar changes.


keghani:
That's strange, this should definitely result in JAR changes, this file is included in the java-build JAR.

Then please run all ant rules and maven to ensure everything compiles, tests pass, and there's no diff in metadata.

keghani:
We investigated this and there is a diff in the JAR file but it's in the non-public implementation of the .class file so doesn't show up in pkgdiff.

@dlyongemallo please include the JARs and other generated files in this PR.

PR #883, opened 2015-11-05
fix bug with nationalPrefixOptionalWhenFormatting being dropped for the US when writing nanoproto files

dlyongemallo:

PR #885, opened 2015-11-06
cherry-pick of googlei18n/libphonenumber#883 into nanoproto branch

dlyongemallo:


keghani:
Did you cherry-pick #883 or #884?

dlyongemallo:
Actually, I did both and verified that their outcome was exactly the same. I don't remember which version I uploaded, but it literally makes no difference.

keghani:
Thanks! That makes sense. 883 tried to do in nanoproto what we decided to do first in master with 884 and follow up with this. The description for ed306d1 said 884 and since that's the PR we merged (and abandoned 883) that's what I used in the description when merging.

Issue #853, opened 2015-10-15
New slovak operator not supported


deel77:
There is new mobile operator (4ka) in Slovakia (SK) and he is using numbers in form of:
0950 xxx xxx
but they are not validated by the library

page of new operator https://www.4ka.sk (only in slovak)

keghani:
Thanks for reporting @deel77 !

We're working on this. I have one question:

4ka is also called SWAN (according to that website), and this is the named used in the national number plan (http://www.itu.int/oth/T02020000BD/en). Is "SWAN" a well-known name locally?

deel77:
SWAN is the company name behind mobile infrastracture, but they partnered with Slovak Post (Slovenská pošta) in order to create mobile operator and they are marketing it as "4ka".

So common people know them as "4ka" but people with technical background should know "SWAN" as well.

keghani:
Thanks! We'll use 4ka.

keghani:
This will be in the next release.

keghani:
This was fixed: http://libphonenumber.appspot.com/phonenumberparser?number=0950+000+000&country=SK

PR #891, opened 2015-11-10
add phone prefix 950 for slovak operator 4ka

peter-veres:
new slovak operator 4ka has prefix 950

for more info see https://www.4ka.sk/tlacove-spravy/a-je-tu-4ka-vas-slovensky-operator-na-slovenskom-trhu


keghani:
Duplicate of #853. This will be in the next release.

PR #928, opened 2015-12-18
add ShortNumberInfo results to Java demo

dlyongemallo:


java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -25,6 +25,7 @@
 import com.google.i18n.phonenumbers.PhoneNumberUtil;
 import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberFormat;
 import com.google.i18n.phonenumbers.PhoneNumberUtil.PhoneNumberType;
+import com.google.i18n.phonenumbers.ShortNumberInfo;
```
lararennie: Sorting off?

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -53,6 +54,7 @@
 @SuppressWarnings("serial")
 public class PhoneNumberParserServlet extends HttpServlet {
   private PhoneNumberUtil phoneUtil = PhoneNumberUtil.getInstance();
+  private ShortNumberInfo shortInfo = ShortNumberInfo.getInstance();
```
lararennie: What's the life of this servlet? once per request?

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -214,6 +217,29 @@ private StringBuilder getOutputForSingleNumber(
       output.append("</TABLE>");
       output.append("</DIV>");
 
+      boolean isPossibleShort = shortInfo.isPossibleShortNumber(number);
```
lararennie: Do we want to do this every time, or only if it isn't a valid number at all? Will we/should we have numbers that are both? Too much output can be confusing. You could instead of this just above have a bit where if it's invalid, put this possible/valid short number info.

dlyongemallo: Won't it be more confusing if short number info shows for some numbers but not others? Okay, I can wrap this in a test for validity.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: As long as it's not garbage collected, I think. Why?

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java

PR #934, opened 2015-12-29
refactoring of demo with bug fixes including XSS fix

dlyongemallo:
1. Use the standard UTF_8 String constant (available since Java 1.7).
2. Set the character encoding for when the input was a file (previously only set when the input was a single input but not a file).
3. Refactor getOutputForSingleNumber so that the HTML generated is self-contained. (Previously, the BODY and HTML closing tags were written outside of the if block for whether it's a single input or a file, leading to the tags being erroneously written twice when it's a file.)
4. Fix a minor bug where whether a newline is printed after the language code was dependent on whether a region code is present.

Edit: 5. Fix an existing bug which allows XSS attack through the inputs.

### DIFF
--- a/java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
+++ b/java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
@@ -18,6 +18,8 @@
 
 package com.google.phonenumbers;
 
+import static java.nio.charset.StandardCharsets.UTF_8;
+
 import com.google.i18n.phonenumbers.AsYouTypeFormatter;
 import com.google.i18n.phonenumbers.NumberParseException;
 import com.google.i18n.phonenumbers.PhoneNumberToCarrierMapper;
@@ -35,6 +37,7 @@ import org.apache.commons.fileupload.FileUploadException;
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
 import org.apache.commons.fileupload.util.Streams;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringEscapeUtils;
 
 import java.io.IOException;
 import java.io.InputStream;
@@ -71,7 +74,7 @@ public class PhoneNumberParserServlet extends HttpServlet {
         if (item.isFormField()) {
           String fieldName = item.getFieldName();
           if (fieldName.equals("phoneNumber")) {
-            phoneNumber = Streams.asString(in, "UTF-8");
+            phoneNumber = Streams.asString(in, UTF_8.name());
           } else if (fieldName.equals("defaultCountry")) {
             defaultCountry = Streams.asString(in).toUpperCase();
           } else if (fieldName.equals("languageCode")) {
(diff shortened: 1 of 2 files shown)


java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -35,6 +37,7 @@
 import org.apache.commons.fileupload.servlet.ServletFileUpload;
 import org.apache.commons.fileupload.util.Streams;
 import org.apache.commons.io.IOUtils;
+import org.apache.commons.lang.StringEscapeUtils;
```
hagbard: For later, also consider:
https://github.com/google/guava/blob/master/guava/src/com/google/common/html/HtmlEscapers.java
The Apache one does entity escaping, which in theory can be trouble (however in this case no expected values are going to cause issues, so that shouldn't be an issue).

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -171,6 +162,17 @@ private void appendLine(String title, String data, StringBuilder output) {
...
+    output.append("<HTML><HEAD>");
+    output.append(
+        "<link type=\"text/css\" rel=\"stylesheet\" href=\"/stylesheets/main.css\" />");
+    output.append("</HEAD>");
+    output.append("<BODY>");
+    output.append("Phone Number entered: " + StringEscapeUtils.escapeHtml(phoneNumber) + "<br>");
+    output.append("defaultCountry entered: " + StringEscapeUtils.escapeHtml(defaultCountry)
+        + "<br>");
+    output.append("Language entered: " + StringEscapeUtils.escapeHtml(languageCode) +
+        (regionCode.length() == 0 ? "" : " (" + StringEscapeUtils.escapeHtml(regionCode) + ")")
```
hagbard: regionCode.isEmpty() - as we've already burned the 1.6 bridge.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -322,6 +324,7 @@ private StringBuilder getOutputForSingleNumber(
     } catch (NumberParseException e) {
       output.append(e.toString());
```
hagbard: This is wrong because the error string can appear in the middle of the output (and may also contain inadvertent metacharacters). Change this so that the NumberParseException can only happen _before_ we do any output writing. Then remove any try/catch from the output code (so any errors just kill the request).

dlyongemallo: This was by design. It was intended to display the input plus the error.

Before:
before

After:
after

Killing the request would remove the information about the error which is currently displayed.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: I didn't want to add a GWT or Guava dependency. There's already a dependency on org.apache.commons.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: The error string comes from a fixed set of strings, btw, so there's no danger of something unexpected there.

dlyongemallo: As discussed, also escaped the exception message since it's output to HTML. In fact, added escaping even to output of libphonenumber when the result is printed to HTML.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -134,7 +125,8 @@ private StringBuilder getOutputForFile(String defaultCountry, String fileContent
       phoneNumberId++;
       output.append("<TR>");
       output.append("<TD align=center>").append(phoneNumberId).append(" </TD> \n");
```
dlyongemallo: This shouldn't need escaping since it's an integer.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -171,6 +166,17 @@ private void appendLine(String title, String data, StringBuilder output) {
   private StringBuilder getOutputForSingleNumber(
       String phoneNumber, String defaultCountry, String languageCode, String regionCode) {
     StringBuilder output = new StringBuilder();
+    output.append("<HTML><HEAD>");
+    output.append(
+        "<link type=\"text/css\" rel=\"stylesheet\" href=\"/stylesheets/main.css\" />");
```
hagbard: uppercase for consistency?

hagbard:
Other than the existing comments, LGTM. Merge when you've done them.

PR #935, opened 2015-12-29
add permalink from result page to come back to that result

dlyongemallo:
~~Note that this is on top of PR #934.~~

Merged.

### DIFF
--- a/java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
+++ b/java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
@@ -19,6 +19,7 @@
 package com.google.phonenumbers;
 
 import static java.nio.charset.StandardCharsets.UTF_8;
+import static java.util.Locale.ENGLISH;
 
 import com.google.i18n.phonenumbers.AsYouTypeFormatter;
 import com.google.i18n.phonenumbers.NumberParseException;
@@ -103,16 +104,40 @@ public class PhoneNumberParserServlet extends HttpServlet {
     resp.setContentType("text/html");
     resp.setCharacterEncoding(UTF_8.name());
     if (fileContents.length() == 0) {
-      resp.getWriter().println(
-          getOutputForSingleNumber(phoneNumber, defaultCountry, languageCode, regionCode));
+      // Redirect to a URL with the given input encoded in the query parameters.
+      Locale geocodingLocale = new Locale(languageCode, regionCode);
+      resp.sendRedirect(getPermaLinkURL(phoneNumber, defaultCountry, geocodingLocale));
     } else {
       resp.getWriter().println(getOutputForFile(defaultCountry, fileContents));
     }
   }
 
+  /**
+   * Handle the get request to get information about a number based on query parameters.
+   */
+  public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    String phoneNumber = req.getParameter("number");
+    if (phoneNumber == null) {
+      phoneNumber = "";
+    }
+    String defaultCountry = req.getParameter("country");
+    if (defaultCountry == null) {
+      defaultCountry = "";
+    }
+    String geocodingParam = req.getParameter("geocodingLocale");
+    Locale geocodingLocale;
+    if (geocodingParam == null) {
+      geocodingLocale = ENGLISH;  // Default languageCode to English if nothing is entered.
+    } else {
+      geocodingLocale = Locale.forLanguageTag(geocodingParam);
+    }
+    resp.getWriter().println(
+        getOutputForSingleNumber(phoneNumber, defaultCountry, geocodingLocale));
+  }
+
   private StringBuilder getOutputForFile(String defaultCountry, String fileContents) {
-    StringBuilder output = new StringBuilder();
-    output.append("<HTML><HEAD><TITLE>Results generated from phone numbers in the file provided:"
+    StringBuilder output = new StringBuilder(
+        "<HTML><HEAD><TITLE>Results generated from phone numbers in the file provided:"
         + "</TITLE></HEAD><BODY>");
     output.append("<TABLE align=center border=1>");
     output.append("<TH align=center>ID</TH>");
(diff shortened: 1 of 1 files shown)


java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -103,13 +103,46 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
     resp.setContentType("text/html");
     resp.setCharacterEncoding(UTF_8.name());
     if (fileContents.length() == 0) {
-      resp.getWriter().println(
-          getOutputForSingleNumber(phoneNumber, defaultCountry, languageCode, regionCode));
+      // Redirect to a URL with the given input encoded in the query parameters.
+      try {
+        resp.sendRedirect(getPermaLinkURL(phoneNumber, defaultCountry, languageCode, regionCode));
+      } catch(UnsupportedEncodingException e) {
+        resp.getWriter().println(StringEscapeUtils.escapeHtml(e.toString()));
```
hagbard: This is unreachable code (UTF-8 is guaranteed in Java). You should just add a comment here and make this throw:
  new AssertionError(e);

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -103,13 +103,46 @@ public void doPost(HttpServletRequest req, HttpServletResponse resp) throws IOEx
...
+  public void doGet(HttpServletRequest req, HttpServletResponse resp) throws IOException {
+    String phoneNumber = req.getParameter("number");
+    if (phoneNumber == null) {
+      phoneNumber = "";
+    }
+    String defaultCountry = req.getParameter("country");
+    if (defaultCountry == null) {
+      defaultCountry = "";
+    }
+    String geocodingLocale = req.getParameter("geocodingLocale");
```
hagbard: Why not use Locale.forLanguageTag() ?

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -161,9 +194,31 @@ private void appendLine(String title, String data, StringBuilder output) {
   }
 
   /**
+   * Returns a stable URL pointing to the result page for the given input.
+   */
+  private String getPermaLinkURL(
+      String phoneNumber, String defaultCountry, String languageCode, String regionCode)
+      throws UnsupportedEncodingException {
+    StringBuilder permaLink = new StringBuilder();
+    permaLink.append("http://libphonenumber.appspot.com/phonenumberparser");
```
hagbard: Just make this a relative URL. It's not actually the "permalink" because it just makes a redirect. We just want the user to end up on a bookmarkable page.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -161,9 +194,31 @@ private void appendLine(String title, String data, StringBuilder output) {
   }
 
   /**
+   * Returns a stable URL pointing to the result page for the given input.
+   */
+  private String getPermaLinkURL(
```
hagbard: I'd move the try/catch in here to prevent leaking the (impossible) exception.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -161,9 +194,31 @@ private void appendLine(String title, String data, StringBuilder output) {
...
+      String phoneNumber, String defaultCountry, String languageCode, String regionCode)
+      throws UnsupportedEncodingException {
+    StringBuilder permaLink = new StringBuilder();
+    permaLink.append("http://libphonenumber.appspot.com/phonenumberparser");
+    permaLink.append("?number=" + URLEncoder.encode(phoneNumber, UTF_8.name()));
+    if (!defaultCountry.isEmpty()) {
+      permaLink.append("&country=" + URLEncoder.encode(defaultCountry, UTF_8.name()));
+    }
+    if (!languageCode.equals("en") || !regionCode.isEmpty()) {
```
hagbard: just use local

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -188,6 +243,9 @@ private String getNewIssueLink(String phoneNumber, String defaultCountry)
         "Supporting evidence (for example, national numbering plan, announcement from mobile "
         + "carrier, news article): **IMPORTANT - anything posted here is made public. "
         + "Read the guidelines first!** \n\n");
+    issueTemplate.append("[link to demo]("
```
hagbard: Do we even want this? We just want the page on which results are shown to be bookmarkable. No need to add any more UI elements or anything.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: I've moved the exception below.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: Good point.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
```diff
@@ -212,6 +273,9 @@ private StringBuilder getOutputForSingleNumber(
     output.append("Language entered: " + StringEscapeUtils.escapeHtml(languageCode) +
         (regionCode.isEmpty() ? "" : " (" + StringEscapeUtils.escapeHtml(regionCode) + ")")
         + "<BR>");
+    output.append("[<a href=\""
+        + getPermaLinkURL(phoneNumber, defaultCountry, languageCode, regionCode)
+        + "\">link to this result page</a>]<br>");
```
dlyongemallo: @hagbard - Any opinion on this link? @roes thinks we should remove it since it'll already be in the URL.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: I can't make it relative as it has to work from github. (That is, unless we don't link back from the github issue.)

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: This is linking from a github issue back to the demo result page for the number which the issue is about. That should be really useful for us.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: I think your above comment is actually about this link? I'll remove it.

java/demo/src/com/google/phonenumbers/PhoneNumberParserServlet.java
dlyongemallo: I've changed the code to use Locale throughout.

PR #965, opened 2016-02-08
fix two places in demo where NPEs can occur on bad input

dlyongemallo:


java/release_notes.txt
```diff
@@ -1,3 +1,6 @@
+Pending changes:
+ - fix two places in demo where NPEs can occur on bad input
```
keghani: s/fix/Fix

keghani: Also please add "Code changes:" between "Pending changes" and the bullet (to distinguish from Metadata changes)

dlyongemallo: Done.

PR #975, opened 2016-02-12
Fixing cmake issues on mac

lararennie:
Two issues:
a) add_custom_target test is a reserved keyword - changed to tests
b) there has been a policy change regarding RPATH. I could explicitly set the policy, but I think it's reasonable to just bump the version of cmake used. Chose 3.1 since it's sufficient but older.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 # Author: Philippe Liard
 
-cmake_minimum_required (VERSION 2.8)
+cmake_minimum_required (VERSION 3.1)
 
 project (libphonenumber)
 set (libphonenumber_VERSION_MAJOR 7)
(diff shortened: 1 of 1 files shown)


lararennie:
(Actually this is one commit. I don't know how to squash the master merge commit - when I use rebase it gives me neither of these as an option.)

keghani:
To squash - git commit -a --amend

This actually is 2 commits but 1 PR :) It's better though in general in my opinion since we can see the diff if you change anything later (so we don't review what was already done).

dlyongemallo:
Squash won't work here: the first commit is not the parent of the second. What needs to be done is that, instead of merging 'master' into makelists-fix, to reset the makelists-fix branch to the latest commit in master, and then roll the "Fixing cmake issues:" commit on top.

git makelists-fix
git reset --hard 3c34928 [or whatever is the latest commit on master]
git rebase -i
pick 0fac8c2 [replace the "noop" line with this]

You should now be one commit ahead of master. You can "git push --force" to replace the contents of your makelists-fix branch with this.

PR #999, opened 2016-03-10
Mark method CountryCodeToRegionCodeMap#getCountryCodeToRegionCodeMap as public

vbauer:
In our project we used the following workaround to get access to the `CountryCodeToRegionCodeMap#getCountryCodeToRegionCodeMap` method:

```java
package com.google.i18n.phonenumbers;

import java.util.List;
import java.util.Map;
import java.util.Set;

public abstract class CountryCodeToRegionCodeMapWrapper {
    static final Map<Integer, List<String>> _CountryCodeToRegionCodeMap = CountryCodeToRegionCodeMap.getCountryCodeToRegionCodeMap();

    public static Map<Integer, List<String>> getCountryCodeToRegionCodeMap(){
        return _CountryCodeToRegionCodeMap;
    }
}
```

We put `CountryCodeToRegionCodeMapWrapper` class in the package `com.google.i18n.phonenumbers` to have access to this method. It will be very useful, to have direct access to this method without above hack.

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/CountryCodeToRegionCodeMap.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/CountryCodeToRegionCodeMap.java
@@ -26,11 +26,15 @@ import java.util.List;
 import java.util.Map;
 
 public class CountryCodeToRegionCodeMap {
-  // A mapping from a country code to the region codes which denote the
-  // country/region represented by that country code. In the case of multiple
-  // countries sharing a calling code, such as the NANPA countries, the one
-  // indicated with "isMainCountryForCode" in the metadata should be first.
-  static Map<Integer, List<String>> getCountryCodeToRegionCodeMap() {
+  /**
+   * A mapping from a country code to the region codes which denote the
+   * country/region represented by that country code. In the case of multiple
+   * countries sharing a calling code, such as the NANPA countries, the one
+   * indicated with "isMainCountryForCode" in the metadata should be first.
+   *
+   * @return mapping from a country code to the region codes
+   */
+  public static Map<Integer, List<String>> getCountryCodeToRegionCodeMap() {
     // The capacity is set to 285 as there are 214 different entries,
     // and this offers a load factor of roughly 0.75.
     Map<Integer, List<String>> countryCodeToRegionCodeMap =


xvik:
+1

zherebjov:
+1

asavinovskikh:
+1

volkovma:
+1

dsagomonov:
+1

penmetsaa:
Can you please let us know why you need this CountryCodeToRegionCodeMap?

vbauer:
@padmaksha I try to describe all our use-cases. We'd like to have access to the following data/operations:

* Get all country codes (`CountryCodeToRegionCodeMap.getCountryCodeToRegionCodeMap().keySet()`)
* Get region codes by country code (I think we can do it via `PhoneNumberUtils#getRegionCodesForCountryCode`, so it would be ok, but we can minimize memory usage a little bit, if we use `CountryCodeToRegionCodeMap` directly without `PhoneNumberUtils`).
* Retrive `CountryCodeToRegionCodeMap.getCountryCodeToRegionCodeMap()` and send this information via GWT RPC on the client side. It is necessary for our UI components.

penmetsaa:
@vbauer

CountryCodeToRegionCodeMap is one of our helper API that is developed for internal API usage. CountryCodeToRegionCodeMap returns the country codes and region codes that libphonenumber supports. Ex: It does not have details for countries like South Georgia & South Sandwich Islands, Antarctica..etc. Exposing helper APIs is not a good thing to do for lot of reasons like maintenance, compatibility.. etc

Can you please specify the exact use case for which you need this map in front end?
Ex: Depending on the country code entered in the UI do you want to get the region code list..etc

Please check other methods like ```PhoneNumberUtil.getRegionCodesForCountryCode(), PhoneNumberUtil.getCountryCodeForRegion()``` if they suits your use case. Or you need to add more logic to create your custom map. Similar APIs are available in [our JS library](https://github.com/googlei18n/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js#L2937) so that you can use directly in your UI components.

penmetsaa:
Please reopen if there is something to mention here.
Thanks for all the support.

lararennie:
@vbauer:
For the specific use-cases that you mention:
1) Get all country codes (CountryCodeToRegionCodeMap.getCountryCodeToRegionCodeMap().keySet() -- what do you use this for? We could theoretically provide a method e.g. getValidCountryCallingCodes() to return a set of all the valid ones if there's a need for this? This would include global network calling codes e.g. +800 -> would you want that or not?

2) Get region codes by country code (I think we can do it via PhoneNumberUtils#getRegionCodesForCountryCode, so it would be ok, but we can minimize memory usage a little bit, if we use CountryCodeToRegionCodeMap directly without PhoneNumberUtils). -- Do you do something other than just parse the number? Is that why you need this? Maybe an image of your widget would help here?

3) Retrieve CountryCodeToRegionCodeMap.getCountryCodeToRegionCodeMap() and send this information from server-side to client side (via GWT RPC). It is necessary for our UI components. -- What do you do with this?

Given the current API, what we'd expect users to do is call getSupportedRegions (and optionally getSupportedGlobalNetworkCallingCodes as well), and then use this to build a custom map of the data they want, keyed off whatever they want. (E.g. you could imagine building a map of calling code to a "country" object that contains an image link to a flag and the name(s) of the countries to show on hover). Does this not work for your use-case? I wouldn't think the region-codes themselves are something you want to show to users directly in your UI.

PR #1006, opened 2016-03-15
Update FALSEHOODS.md

kylecronin:
Adding additional falsehoods. Feel free to take them or leave them, but I have encountered programmers that believe them.

### DIFF
--- a/FALSEHOODS.md
+++ b/FALSEHOODS.md
@@ -5,16 +5,23 @@
 Given how ubiquitous phone numbers are and how long they've been around, it's
 surprising how many false assumptions programmers continue to make about them.
 
+1. **An individual has a phone number**
+
+    Some people do not own phones, or do not wish to provide you with their
+    telephone number when asked. Do not require a user to provide a phone number
+    unless it is essential, and whenever possible try and provide a fallback
+    to accommodate these users.
+
+1.  **An individual has only one phone number**
+
+    Obviously, this isn't necessarily true.
+
 1.  **A phone number uniquely identifies an individual**
 
     It wasn't even that long ago that mobile phones didn't exist, and it was
     common for an entire household to share one fixed-line telephone number. In
     some parts of the world, this is still true, and relatives (or even friends)
     share a single phone number.
-
-1.  **An individual has only one phone number**
-
-    Obviously, this isn't necessarily true.
 
 1.  **Phone numbers cannot be re-used**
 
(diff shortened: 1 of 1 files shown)


FALSEHOODS.md
```diff
@@ -5,6 +5,13 @@
 Given how ubiquitous phone numbers are and how long they've been around, it's
 surprising how many false assumptions programmers continue to make about them.
 
+1. **An individual has a phone number**
```
lararennie: I think this is a good point, and a good thing to start with. We should probably though make the "An individual has only one phone number" come straight after it, and move the "Phone numbers that are valid today" one down to after "Phone numbers cannot be re-used".

If you would like to do the reordering, feel free to do so and upload another patch, otherwise we'll do a separate PR with this new falsehood and the reordering in it.

FALSEHOODS.md
```diff
@@ -53,6 +60,14 @@ surprising how many false assumptions programmers continue to make about them.
     from within a subset of countries, such as the international 00800 numbers.
     Some may be dialable only if the caller is a subscriber to a particular
     telecom company.
+    
+1. **You can send a text message to any phone number**
+
+    A lot of people still only have a home telephone, which typically cannot
```
lararennie: home -> fixed-line

We should separate the info from the "tip" part, like we did above with "Each country has only one country calling code".

FALSEHOODS.md
```diff
@@ -53,6 +60,14 @@ surprising how many false assumptions programmers continue to make about them.
...
+    send or receive text messages. Whenever possible try and accommodate these
+    users by calling them with a voice message or voice prompts. Alternatively,
+    if your service requires sending text messages, verify that the number
+    can receive texts messages after the user provides it.
```
lararennie: s/texts/text

FALSEHOODS.md
```diff
@@ -53,6 +60,14 @@ surprising how many false assumptions programmers continue to make about them.
     from within a subset of countries, such as the international 00800 numbers.
     Some may be dialable only if the caller is a subscriber to a particular
     telecom company.
+    
+1. **You can send a text message to any phone number**
```
lararennie: We should probably add the corollary: that "Only mobile phones can receive text messages" which is also not true.

kylecronin:
I signed it!

FALSEHOODS.md
kylecronin: I updated this PR with the reordering

FALSEHOODS.md
kylecronin: Done

FALSEHOODS.md
kylecronin: Good catch, fixed

FALSEHOODS.md
kylecronin: Done, section added

lararennie:
Thanks so much for your contribution!!

PR #1009, opened 2016-03-15
Adding one more falsehood, from historical example.

myrkraverk:


FALSEHOODS.md
```diff
@@ -118,3 +118,8 @@ surprising how many false assumptions programmers continue to make about them.
 
     In Egypt, it is common for phone numbers to be written in native digits.
 
+1.  **Phone numbers have only one prefix**
```
lararennie: By prefix, do you mean national destination code (or "area code" for geographical numbers)?

Maybe a better title would be "There is only one sequence of digits that reaches a single end-point"? Although that isn't so readable by normal people. Maybe: "There is only one area code or national destination code for a given phone number at a given time"? Or "Phone numbers have only one area code or national destination code at a given time"?

FALSEHOODS.md
```diff
@@ -118,3 +118,8 @@ surprising how many false assumptions programmers continue to make about them.
...
+
+    In the mid-90s in Iceland, phone numbers changed from 5 and 6 digits to 7 digits.
```
lararennie: I'd start off saying "It is common during transition periods for countries to support both old and new ways of dialling a phone number." And then you can add the Iceland example.

FALSEHOODS.md
```diff
@@ -118,3 +118,8 @@ surprising how many false assumptions programmers continue to make about them.
...
+    During the transition period, phone numbers could be reached by the old area code
+    or the new 7 digit number (a different prefix).
```
lararennie: Do we have any concrete tips on what this means for programmers - i.e. what they should/shouldn't do because of this?

myrkraverk: According to an old news article (in Icelandic) the transition begun in one region before the others (I can't verify this anymore) so applying automated updates to entire databases is not a good idea; at least not during the transition period.  It also means if you're looking for a record by phone number, you have to consider both prefixes.

If you are writing an auto dialing tool for, say helpdesk operators, you have to be aware of which region is available through the new seven digit number, or the old regional code, or both.

FALSEHOODS.md
myrkraverk: The old system had regional prefixes.  The new system has seven digit phone numbers and no regional codes.  So during the transition, a Reykjavik phone number could be dialed with nnnnn and 55nnnnn inside the region; and 91-nnnnn and 55nnnnn from outside.

lararennie: Ah I see, so the actual prefix changed. Can you add this example into the text?

FALSEHOODS.md
lararennie: Okay, so we could add something like:

**Tip:** During transition periods, make sure that both forms of a number are supported for look-up by phone number. If sending messages/dialling affected numbers, both the new and old prefix may need to be tried. Depending on the type of transition, you may need to contact affected customers and ask them to update their number.

lararennie:
Thanks so much for the contribution!

PR #1014, opened 2016-03-16
Adding ifdef I18N_PHONENUMBERS_USE_ALTERNATE_FORMATS in phonenumbermatcher_test

lararennie:
Fixed issue with failing tests for users who choose not to use alternate formats.

Added "ifdef" around the entries in phonenumbermatcher_test that require alternate formats to be included to be found. This means that if the user doesn't have this option set to true, the tests won't be run (they would otherwise fail).


andyst:
Where is I18N_PHONENUMBERS_USE_ALTERNATE_FORMATS documented?
I don't see it defined anywhere, I don't see it mentioned in the README (global or cpp).

lararennie:
It is in the MakeLists file declared as "USE_ALTERNATE_FORMATS" - one of the many options that developers can configure.

It is in the README file as a "supported build parameter".

Specifying this in ccmake causes I18N_PHONENUMBERS_USE_ALTERNATE_FORMATS to be set (in CMakeLists).

PR #1015, opened 2016-03-16
Add another example of * in phone numbers

steveh:
Public supporting evidence: http://www.police.govt.nz/advice/driving-and-road-safety/report-bad-driver

### DIFF
--- a/FALSEHOODS.md
+++ b/FALSEHOODS.md
@@ -112,7 +112,7 @@ surprising how many false assumptions programmers continue to make about them.
 
 1.  **Phone numbers contain only digits**
 
-    In Israel, certain advertising numbers start with a `*`.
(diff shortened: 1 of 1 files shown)


FALSEHOODS.md
```diff
@@ -112,7 +112,7 @@ surprising how many false assumptions programmers continue to make about them.
 
 1.  **Phone numbers contain only digits**
 
-    In Israel, certain advertising numbers start with a `*`.
+    In Israel, certain advertising numbers start with a `*`. In New Zealand, non-urgent traffic incidents can be reported by calling      `*555` from a mobile phone.
```
lararennie: If you can just fix the alignment here and remove the extra spaces after "calling" we can just merge this in directly. Thanks!

PR #1016, opened 2016-03-17
Update URLs to files and directories in README

ifraixedes:


lararennie:
Thanks so much for this!

PR #1027, opened 2016-03-20
FALSEHOODS.md: Add two more falsehoods

joshtriplett:


FALSEHOODS.md
```diff
@@ -118,3 +118,18 @@ surprising how many false assumptions programmers continue to make about them.
...
 
+1.  A given phone number has only one associated phone line. (Corollary: if a
+    call is ongoing with a given phone number, another call cannot be made to
+    or from that number until the first call finishes.)
+
+    In addition to services like call waiting, many phone services (especially
+    for businesses) allow multiple outbound calls from the same phone number.
+
+1.  The area code or country code of a phone number gives some idea of the
+    user's location, time zone, or preferred language.
```
lararennie: I think it is actually fair it gives you some idea. If they have an American phone number, statistically speaking, it's more likely that they are in the US than somewhere else, and that they speak English.

I get your point though that it shouldn't be assumed that it is accurate.

Would you be happier with this more definite sentence instead: "The country or area code of a phone number indicates the user's location, time-zone or preferred language."

We could also add "place of residence" to the list.

FALSEHOODS.md
```diff
@@ -118,3 +118,18 @@ surprising how many false assumptions programmers continue to make about them.
...
+
+    For area codes, see <https://xkcd.com/1129/>: mobile phones and number
+    portability make it entirely likely that someone will have an area code
+    from another state or region.  Similarly, international travelers may
+    have a number from another country.  And VoIP services make it possible
```
lararennie: Remove And.

Also add a point that there are many people who choose to have a mobile phone number that is registered to another country, e.g. people in Liechtenstein may have a Swiss mobile number, if many of their contacts are in Switzerland.

(It's not just travellers.)

FALSEHOODS.md
```diff
@@ -118,3 +118,18 @@ surprising how many false assumptions programmers continue to make about them.
...
+    for users to use a phone number from anywhere in the world.
```
lararennie: We try and make concrete suggestions where we can for what programmers should do because of this, so a tip we could say here is that programmers should therefore make sure not to disallow users from signing up with a phone number from another country, and should be very careful when using methods like isValidNumberForRegion.

FALSEHOODS.md
```diff
@@ -118,3 +118,18 @@ surprising how many false assumptions programmers continue to make about them.
 
     In Egypt, it is common for phone numbers to be written in native digits.
 
+1.  A given phone number has only one associated phone line. (Corollary: if a
```
lararennie: I'm not sure how I understand how this assumption could cause programmers to make mistakes: what could they be doing wrong if they wrongly assume this?

FALSEHOODS.md
```diff
@@ -118,3 +118,18 @@ surprising how many false assumptions programmers continue to make about them.
...
+
+    In addition to services like call waiting, many phone services (especially
+    for businesses) allow multiple outbound calls from the same phone number.
+
+1.  The area code or country code of a phone number gives some idea of the
+    user's location, time zone, or preferred language.
+
+    For area codes, see <https://xkcd.com/1129/>: mobile phones and number
+    portability make it entirely likely that someone will have an area code
+    from another state or region.  Similarly, international travelers may
```
lararennie: Perhaps also worth mentioning that in some countries, not only fixed-line phones can be associated with geographical area codes. For users outside the Americas, this can be surprising that they are at all issued that way.

PR #1034, opened 2016-03-21
Add a note about the Italian leading zero

Flameeyes:


FALSEHOODS.md
```diff
@@ -118,3 +118,9 @@ surprising how many false assumptions programmers continue to make about them.
 
     In Egypt, it is common for phone numbers to be written in native digits.
 
+1.  **The leading zero in domestic numbers can always be discarded when dialing
```
lararennie: This is a good one! Maybe we should order it after "To make a number dialable, you only need to change the prefix" since it's also about dialling prefixes?

Also, perhaps substitute "domestic numbers" with "numbers formatted for domestic usage" and change "The leading zero" to "A leading zero" since there may not be one?

FALSEHOODS.md
```diff
@@ -118,3 +118,9 @@ surprising how many false assumptions programmers continue to make about them.
...
+    internationally.**
+
+    In Italy, since 1998, the prefix was "fixed" onto the phone numbers, so
+    (01) 2345 became 012345 and should be dialed (internationally) as
+    +39012345 (including the leading zero.
```
lararennie: a) missing a trailing bracket
b) optional: could mention that Italy had a plan to change this, but it hasn't eventuated.
c) could add a "Tip" that people should use the library's parse and format numbers to handle this correctly.

Issue #822, opened 2015-09-17
Multiple undefined symbols 


PeterBocan:
Hello guys, I'm trying to build library on OS X Yosemite (10.10.5, Boost 1.58, ICU 55). I did run this command:
    
    cmake -DGTEST_INCLUDE_DIR:PATH=/Users/pb1/Downloads/gtest/include/ \
    -DGTEST_SOURCE_DIR:PATH=/Users/pb1/Downloads/gtest ..

Which resulted into some weird issue: It has successfully built static library for libgeocoding.a but it did fail compiling dynamic library (libgeocoding.dylib) in linking stage:
   
    Undefined symbols for architecture x86_64:
    "i18n::phonenumbers::SimpleItoa(int)", referenced from:
        i18n::phonenumbers::AreaCodeMap::Lookup(i18n::phonenumbers::PhoneNumber const&) const in area_code_map.cc.o
    "i18n::phonenumbers::SimpleItoa(long long)", referenced from:
      i18n::phonenumbers::AreaCodeMap::Lookup(i18n::phonenumbers::PhoneNumber const&) const in area_code_map.cc.o
    "i18n::phonenumbers::safe_strto64(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&, long long*)", referenced from:
      i18n::phonenumbers::AreaCodeMap::Lookup(i18n::phonenumbers::PhoneNumber const&) const in area_code_map.cc.o
    "i18n::phonenumbers::PhoneNumberUtil::GetInstance()", referenced from:
      i18n::phonenumbers::AreaCodeMap::AreaCodeMap() in area_code_map.cc.o
      i18n::phonenumbers::PhoneNumberOfflineGeocoder::Init(int const*, int, i18n::phonenumbers::CountryLanguages const* (*)(int), char const**, int, i18n::phonenumbers::PrefixDescriptions const* (*)(int)) in phonenumber_offline_geocoder.cc.o
    "i18n::phonenumbers::PhoneNumberUtil::kRegionCodeForNonGeoEntity", referenced from:
      i18n::phonenumbers::PhoneNumberOfflineGeocoder::GetRegionDisplayName(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const*, icu_55::Locale const&) const in phonenumber_offline_geocoder.cc.o
[...]

PR #1036, opened 2016-03-23
Add cmake Mac flag to enable Linux builds at lower cmake versions.

keghani:
This is similar to the fix of #822.

Also update README to cover Mac installation.

This rolls back the cmake minimum version bump in #975 since it was
backwards-incompatible, and the following #984. Not expected to cause
any new user-facing regressions, but expected to solve any that were
introduced with #975.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 # Author: Philippe Liard
 
-cmake_minimum_required (VERSION 3.1)
+cmake_minimum_required (VERSION 2.8)
 
 project (libphonenumber)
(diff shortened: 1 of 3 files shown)


keghani:
I'll need to update pending changes as well.

keghani:
Please don't review for now, I need to revise something.

keghani:
Closing, I'll send another PR.

keghani:
Here's the original commit with @andyst's comments: https://github.com/googlei18n/libphonenumber/commit/f0eaed045c0c8b14192aafd9583a7b4b2540d13a

PR #1050, opened 2016-03-31
Parse documentation and exampleNumber updates

lararennie:
* New useful getExampleNumber methods (for testing) - one for invalid
  numbers and one for getting a number of a particular type without
  specifying a country.


keghani:
Looks good, thanks!
How about https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/test/com/google/i18n/phonenumbers/ExampleNumbersTest.java? You can also send a separate PR rather than squashing/amending this.

lararennie:
Good point! For some reason I thought that wasn't open-sourced. Adding it to this one.

keghani:
Thanks!

PR #1052, opened 2016-03-31
Add cmake Mac flag to enable Linux builds at lower cmake versions.

keghani:
This is similar to the fix of #822.

Also update README to cover Mac installation.

This rolls back the cmake minimum version bump in #975 since it was
backwards-incompatible, and the following #984. Not expected to cause
any new user-facing regressions, but expected to solve any that were
introduced with #975.

### DIFF
--- a/cpp/README
+++ b/cpp/README
@@ -56,7 +56,7 @@ Requirements:
     recent Debian-based GNU/Linux distributions).
 
     You can check which version is available:
-    $ apt-cache show libprotobuf-dev 
+    $ apt-cache show libprotobuf-dev
     Package: libprotobuf-dev
     Source: protobuf
     Version: 2.5.0-9ubuntu1   <-- This must be >= 2.4.0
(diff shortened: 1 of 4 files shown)


keghani:
@lararennie this is a second review. @andyst took a look in https://github.com/googlei18n/libphonenumber/pull/1036.

FYI @roubert 

Tested:
- Linux with cmake 3.2.2
- Linux with cmake 2.8.12.2
- Mac with cmake 3.5.1

lararennie:
This is okay too, I guess we're doing this until the min cmake version available on ubuntu etc. is higher? It would be nicer to just bump the version, I considered setting this at the time but decided version bumping was cleaner. But I'm happy to be persuaded otherwise if it makes installation difficult :)

While you are here, our goobuntu instructions (tested today) don't work, we need to add protoc to the apt-get list.... protobuf-compiler needs adding at line 37 in README

keghani:
Yes, we're doing this until the package manager on Ubuntu can give us higher versions, to avoid contributors having to install from source. It's difficult to automate this :)

Re. goobuntu instructions - protoc is more complicated, I can't reproduce right now since I have the "correct" a.k.a. installed-from-source version when testing nanoproto. I'd want to clean that up properly too so everyone is minimally disrupted, but this flag is blocking @padmaksha from releasing.

keghani:
I did make the protobuf-compiler change - I'd misunderstood what you meant ealier :)
Also merged latest changes from master.

PTAL

cpp/README
```diff
@@ -138,6 +139,63 @@ Building and testing the library
...
+Get the libphonenumber source. For example:
+  $ mkdir ~/libphonenumber_clone
+  $ cd ~/libphonenumber_clone
+  $ git clone https://github.com/googlei18n/libphonenumber.git
+
+Build and test the library:
+  $ cd libphonenumber/cpp
+  $ mkdir build
+  $ cd build
+  Replace XXX with the appropriate version number:
```
lararennie: Maybe "Replace XXX in the commands below with the appropriate version number:"

lararennie:
One nit, otherwise LGTM

keghani:
Done. Thanks!

Issue #181, opened 2014-12-03
Use animal-sniffer to ensure Java 1.5 compatibility


mihaelacr-google:
Imported from [Google Code issue #181](https://code.google.com/p/libphonenumber/issues/detail?id=181) created by [candrewsintegralblue](https://code.google.com/u/114215788913124607468/) on 2012-08-19T03:59:46.000Z:

----
libphonenumber seems to strive for Java 1.5 compatibility, and it's really easy to accidentally use something from &gt; Java 1.5 and not find out about it until you actually test on a Java 1.5 VM.

animal-sniffer scans the output bytecode and checks to make only APIs you want to be used are used.

To use it, simply add this block to https://code.google.com/p/libphonenumber/source/browse/trunk/java/pom.xml :

            &lt;plugin&gt;
                &lt;groupId&gt;org.codehaus.mojo&lt;/groupId&gt;
                &lt;artifactId&gt;animal-sniffer-maven-plugin&lt;/artifactId&gt;
                &lt;version&gt;1.8&lt;/version&gt;
				&lt;executions&gt;
					&lt;execution&gt;
					    &lt;phase&gt;test&lt;/phase&gt;
						&lt;goals&gt;
							&lt;goal&gt;check&lt;/goal&gt;
						&lt;/goals&gt;
					&lt;/execution&gt;
				&lt;/executions&gt;
                &lt;configuration&gt;
                    &lt;signature&gt;
                        &lt;groupId&gt;org.codehaus.mojo.signature&lt;/groupId&gt;
                        &lt;artifactId&gt;java15&lt;/artifactId&gt;
                        &lt;version&gt;1.0&lt;/version&gt;
                    &lt;/signature&gt;
                &lt;/configuration&gt;
            &lt;/plugin&gt;

keghani:
Closing as this does not seem to be an issue with the library. Consider joining [the discussion group](https://groups.google.com/forum/#!forum/libphonenumber-discuss).

keghani:
@lararennie corrected me, this might be related to the library :) Reopening.

keghani:
We've accepted this issue and will fix it in an upcoming release.

keghani:
Fixed in 7.4.4, released last week.

PR #1088, opened 2016-04-22
Requiring Java version 1.5 to be used for our APIs.

lararennie:
Demo is exempt from this requirement. animal-sniffer plugin used for this.

See https://github.com/googlei18n/libphonenumber/issues/181


lararennie:
I'm trying to fix the issue with the jars... that is a separate build issue with nanoproto that I am having.

keghani:
In general looks good, but I want to patch it in and do some release dry runs once the build is resolved, on my machine - to make sure you didn't do local setup to get some minor bug to work - which I did once :)

lararennie:
Yep, sounds good.

Note I changed something in phonenumberutil to use isEmpty and it fails as
expected (that is not allowed in 1.5).
Also if I enable it on the demo it fails as expected (we aren't staying
v1.5 there).

2016-04-22 14:23 GMT+02:00 Keghani Kouzoujian <notifications@github.com>:

keghani:
I'll take a look next week given upcoming holidays and deadlines.

java/libphonenumber/src/com/google/i18n/phonenumbers/nano/Phonemetadata.java
```diff
@@ -2,6 +2,8 @@
 
 package com.google.i18n.phonenumbers.nano;
 
+// @@protoc_insertion_point(imports_scope)
```
keghani: This class is not hand-generated.

Did you generate the changes? Or were these added manually?

keghani: Also I can't leave a comment on the jar files - we need to regenerate them before merging (to get the latest from the master merges), and squash the commits.

lararennie: Yes, the jars need to be regenerated.

This class: *no* idea why there is a change. This happened because I ran mvn clean and mvn build.... nothing to do with my cl, but maybe to do with the protoc version I have on my computer.

keghani:
Actually why include the jar files? They're not generated with these poms.

java/libphonenumber/src/com/google/i18n/phonenumbers/nano/Phonemetadata.java
keghani: Ah, that's good (that it was auto-generated) :)

I'll patch it in and see if I get the same output.

lararennie:
I included the jar files because I ran mvn package (to test everything runs okay and tests pass) and then they had a diff so...

But we can revert. We can do a separate update of them at some point if we want to.

keghani:
Yeah I got that diff too when doing the Maven hierarchy fix (https://github.com/googlei18n/libphonenumber/pull/1160), it's the new getExampleNumber API and isNumberGeographical. Let's do them separately (even separately from the Maven hierarchy fix) after this.

keghani:
Also should we add this to the pending changes? It doesn't affect the code, it's just a test for us, right?

lararennie:
Yes, you're right, so I think we don't need to.

2016-06-21 11:43 GMT+02:00 Keghani Kouzoujian <notifications@github.com>:

keghani:
I ran a test in another branch (https://github.com/googlei18n/libphonenumber/commit/69ada95b369ce601e2602359c26765b04d833cba). Running `mvn package` from `java/` results in no changes to tracked files (and none to Phonemetadata.java).

Can you try `mvn package` in a clean setup from `java/`? We should not be changing Phonemetadata.java, I wonder if this will come up for you every time.

lararennie:
Tried it, it only happens when I do it from head, not from java/, but it
happens in a clean set-up too.

I'm using libprotoc 2.6.1 - what are you using? (protoc --version)

2016-06-21 15:10 GMT+02:00 Keghani Kouzoujian <notifications@github.com>:

keghani:
Running `mvn package` from root with these `pom.xml` file changes also gives me no changes in `Phonemetadata.java` (only in the checked-in jar files and some untracked files, as expected).

`protoc --version` gives `libprotoc 2.6.1`, but this may be different than your 2.6.1 (I think installed from sources).
`which protoc` gives `/usr/bin/protoc`
What do you have under `~/.m2/repository/com/google/protobuf/protobuf-java/`? I have just one directory, `2.6.1`.

keghani:
Can we limit the scope here to the 4 pom.xml files? The changes there by themselves don't cause the changes in the jars or Phonemetadata.java. We can investigate that in https://github.com/googlei18n/libphonenumber/pull/1172 and https://github.com/googlei18n/libphonenumber/pull/1170.

The pom.xml files by themselves look good for merging.

lararennie:
Right. Reverted them :)

keghani:
Thanks, LG :)

Issue #824, opened 2015-09-19
Missing dirent.h on Windows 


PeterBocan:
Hello guys, I want to notice you that you are using dirent.h (UNIX library) also on Windows:
[generate_geocoding_data.cc@19](https://github.com/googlei18n/libphonenumber/blob/master/tools/cpp/src/cpp-build/generate_geocoding_data.cc#L19)

This won't work. 
Sincerely, 
Peter Bocan.

pnkjvvdy:
I also can't build libphonenumber with geocoder due to this issue. Is there any another way???

keghani:
This seems infeasible given our bandwidth and expertise, but if you the reader have these then please
feel free to re-open this issue and take it up.

We don't actively maintain windows builds but would be able to accept a PR.
We'd also need someone else in the open-source community to test it out and help with the review. Please see the [known Windows issues](FAQ.md#what-about-windows), maybe someone else there or in the
[discussion group](https://groups.google.com/group/libphonenumber-discuss) could help.

PeterBocan:
@keghani I may one day push PR when C++17 is stable (upcoming filesystem implementation).

keghani:
Thanks @PeterBocan! Feel free to re-open then.

PR #1090, opened 2016-04-23
Replace POSIX directory operations by Boost Filesystem

gjasny:
To support compiling the generate_geocoding_data generator on non-POSIX systems like Windows I replaced the POSIX code with Boost Filesystem. After the change the generator produes exactly the same output as before. This should close Issue #824.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1090)


andyst:
@hagbard @roubert for added C++ and Windows (kidding) expertise.

cpp/CMakeLists.txt
```diff
@@ -82,14 +82,14 @@ if (${USE_ALTERNATE_FORMATS} STREQUAL "ON")
 endif ()
 
 # Find all the required libraries and programs.
-if (${USE_BOOST} STREQUAL "ON")
+if (${USE_BOOST} STREQUAL "ON" OR ${BUILD_GEOCODER} STREQUAL "ON")
```
andyst: This doesn't look right to me.
We can't replace "USE_BOOST" check with "USE_BOOST OR BUILD_GEOCODER" when the check is guarding includes for Boost headers.

andyst:
Still need to review the rest of the code. Just added an initial comment.

landrix:
How can i prepare Boost for Filesystem?

cmake throws
"Can't find Boost Date_Time/Filesystem/System/Thread: can't locate Boost."

cpp/CMakeLists.txt
andyst: I take this back. In this context this is doing the right thing. My initial reading was wrong.

andyst:
@roubert could you please take a look at this CL in terms of C++?
Looks good to me but maybe you spot something.

tools/cpp/src/cpp-build/generate_geocoding_data.cc
```diff
@@ -31,6 +30,7 @@
 #include <string>
 #include <utility>
 #include <vector>
+#include <boost/filesystem.hpp>
```
andyst: This code needs to continue to work without Boost.

Please follow the pattern in https://github.com/googlei18n/libphonenumber/blob/master/cpp/src/phonenumbers/base/memory/scoped_ptr.h to use Boost "#if defined(I18N_PHONENUMBERS_USE_BOOST)" and fall back to the current code based on dirent.h.

ginggs:
[line 58 of tools/cpp/CMakeLists.txt
](https://github.com/gjasny/libphonenumber/blob/45bd90ab3e910a14a82e3889a0b309dd4157a283/tools/cpp/CMakeLists.txt#L58) should be:

`target_link_libraries (generate_geocoding_data_test ${TEST_LIBS} ${Boost_LIBRARIES})`

instead of:

`target_link_libraries (generate_geocoding_data_test ${TEST_LIBS})`

so that **generate_geocoding_data_test** also gets linked with Boost

keghani:
Hi @gjasny,

Are you interested in / able to help us get this PR merged in the foreseeable future?
Sorry Windows hasn't seen much love lately, but we don't actively maintain Windows builds and couldn't test such a PR.
If someone else in the open-source community (@ginggs? Thanks for helping review too!) could test it out and help with the review we'd accept it.

Related issues in case anyone could help collaborate there:
*   #1555, to allow Windows to build cpp library with pthreads for multi-threading (would this still be an issue after this PR?)
*   #1307, to use readdir instead of readdir_r

keghani:
This is infeasible given the lack of reviewers or a response.

We don't actively maintain windows builds and we'd need someone else in the open-source community to test and help review this PR. Please see the [known Windows issues](https://github.com/googlei18n/libphonenumber/blob/master/FAQ.md#what-about-windows), maybe someone else there or in the
[discussion group](https://groups.google.com/group/libphonenumber-discuss) could help.

PR #1091, opened 2016-04-23
Explicitly call std::stringstream::fail()

gjasny:
Implicit cast to bool does not work with C++11 and later. See http://en.cppreference.com/w/cpp/io/basic_ios/operator_bool


andyst:
Thank you!

PR #1102, opened 2016-05-03
Penmetsaa update 7.3.2

penmetsaa:
May 03, 2016: libphonenumber-7.3.2
Metadata changes:
 - Updated phone metadata for region code(s):
   AU, BR, ET, IN, KH, ML, NO, SB, TJ, US
 - Updated short number metadata for region code(s): CA, ML, US
 - New geocoding data for country calling code(s):
   55 (pt), 223 (en), 1548 (en), 1743 (en)
 - Updated geocoding data for country calling code(s): 27 (en), 55 (en)
 - Updated carrier data for country calling code(s):
   61 (en), 91 (en), 223 (en), 251 (en), 677 (en), 992 (en)

Unit test updates for changes to Canada short number metadata.

PR #1104, opened 2016-05-04
Update 7.3.2 release notes

keghani:
1. Changing release notes back to what was automatically generated:

a. 55_pt was actually new this time (it was removed in a recent release): https://github.com/googlei18n/libphonenumber/commits/master/java/geocoder/src/com/google/i18n/phonenumbers/geocoding/data/55_pt

b. And we've always included the entire 1xxx numbers according to the file names - see libphonenumber-7.1.1 release notes for example.

We should generally leave the "metadata" section as is, and if it's wrong see why it was wrongly generated. Also if updating the release notes remember to update the PR description (https://github.com/googlei18n/libphonenumber/pull/1102 stayed the same).

2. Include the non-metadata change as "code changes", which slipped as it was added during the same PR as the release.

### DIFF
--- a/java/release_notes.txt
+++ b/java/release_notes.txt
@@ -1,11 +1,13 @@
 May 03, 2016: libphonenumber-7.3.2
+Code changes:
+ - Updated unit tests for Canada short numbers.
 Metadata changes:
  - Updated phone metadata for region code(s):
    AU, BR, ET, IN, KH, ML, NO, SB, TJ, US
  - Updated short number metadata for region code(s): CA, ML, US
- - New geocoding data for country calling code(s): 223 (en)
- - Updated geocoding data for country calling code(s): 
-   27 (en), 55 (en, pt), 1 (en)
+ - New geocoding data for country calling code(s):
+   55 (pt), 223 (en), 1548 (en), 1743 (en)
+ - Updated geocoding data for country calling code(s): 27 (en), 55 (en)
  - Updated carrier data for country calling code(s):
    61 (en), 91 (en), 223 (en), 251 (en), 677 (en), 992 (en)
 


lararennie:
Good point about the PR description, sorry.
a) I thought we refer to the text file, not the binary, and the text file was a modification, not an add
b) We have done both. We should ideally fix the script, but it's too much effort. 5.1.3 and 5.9.1 just say 1 (en), and previously we did US and not country calling codes at all.
c) I don't think unit-test fixes count as code changes?

keghani:
a) Well the text file and binary file are supposed to be doing the same thing, so the script is looking at only one of them and I guess we went wrong in release 7.2.4 when we removed the 55_pt binary while the text file was still there: https://github.com/googlei18n/libphonenumber/tree/libphonenumber-7.2.4/resources/geocoding/pt. Will investigate separately. You're right, in this change pt/55.txt was only edited.

b) OK I don't understand the 1 "country calling codes". So 1548 and 1743 are in https://github.com/googlei18n/libphonenumber/blob/master/resources/geocoding/en/1.txt but 1767 isn't and that's supposed to be Dominica (not US), right?
Why is it too much effort to fix in the script? It's also not ideal to remember to look at this manually or understand inconsistent release notes. If it makes it easier we can just say "prefixes" instead of "country calling codes" or something.

c) "Test changes" then?

lararennie:
a) I think best for script to look at txt files, don't remember why we made it look at the binary files instead and whether this is actually an oversight.

b) Everything for the US should be in the 1.txt file. The reason 1767 isn't is because we have nothing more detailed than 1767 itself, which we already know is Dominica without using the OfflineGeocoder at all. When building binary files, we split the US file (for performance reasons IIRC) into 1 per area code, and we have special "if 1" code that handles reading from them in the geocoder. I guess it's not too much effort to fix the script if we change it to look at the .txt files not the binary files, just don't remember why it wasn't like that to start with....

c) Sure, sounds good.

keghani:
a) Probably to account for changes done in PRs :) Maybe 55_pt was added in a PR. b/29443300 to investigate later.

b) Thanks, I understand better. b/29441689

c) I don't have a strong opinion :) Thanks for helping me understand the issues, closing this PR.

PR #1112, opened 2016-05-17
Update FAQ.md

penmetsaa:
To test whether custom hyperlink is working / not in GitHub. So that we can change the M2M question more freely in future.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -60,7 +60,7 @@ Some number ranges are explicitly defined as being for fixed-line or mobile
 phones. We even represent ranges defined as being "Mostly land-line" in this
 way.
 
-### What about M2M (machine to machine) numbers?
+### What about M2M (machine to machine) numbers? {#m2m_faq}
 
 libphonenumber does not support M2M numbers at the moment, but might in the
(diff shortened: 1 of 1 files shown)


dlyongemallo:
Did you mean to make a PR? If so, the title should be more descriptive than "Update FAQ.md".

There already is a direct link to the M2M section:
https://github.com/googlei18n/libphonenumber/blob/master/FAQ.md#what-about-m2m-machine-to-machine-numbers

Your suggested syntax won't work, as you can see here:
https://github.com/googlei18n/libphonenumber/blob/2dd50472f61bbdc9ed483d960e27ddec0a10fc7b/FAQ.md#what-about-m2m-machine-to-machine-numbers-m2m_faq

I'm not sure where you got this idea from? 

You _could_ do something like this:
`### What about M2M (machine to machine) numbers?<a name="m2m_faq"></a>`

But then you'll end up with two hyperlinks to the section header.

keghani:
Please don't merge this - the tag should change at the same time as the content (nothing to fix until b/28156174).

PR #1125, opened 2016-05-24
Make DispatchRE2Call() and its callers use the array-oriented functions in RE2.

junyer:
A future change to RE2 will replace the VariadicFunction2 objects with variadic
templates, which would break the code in libphonenumber as it currently stands.


junyer:
Ugh. I will redo this so that it uses the right email address for the CLA check.

PR #1126, opened 2016-05-24
Make DispatchRE2Call() and its callers use the array-oriented functions in RE2.

junyer:
A future change to RE2 will replace the VariadicFunction2 objects with variadic
templates, which would break the code in libphonenumber as it currently stands.


junyer:
The tests in `libphonenumber_test` passed:

```
…
[----------] Global test environment tear-down
[==========] 301 tests from 15 test cases ran. (6606 ms total)
[  PASSED  ] 301 tests.
```

PR #1144, opened 2016-06-03
Update "What valid means" client FAQ

keghani:


keghani:
This was reviewed separately so I'll merge.

PR #1155, opened 2016-06-11
Iran geocoding

afarsi:
I updated this list based latest version of itu document [2015-8-8]
Note: I removed duplicate lines for North Khorasan and South Khorasan which was in previews edition of 98.txt
"North Khorasan Province" have one code, not three codes. I removed another duplicate two lines. Also for "south khorasan" this changes applied.
These changes are based on latest itu document.

### DIFF
--- a/resources/geocoding/en/98.txt
+++ b/resources/geocoding/en/98.txt
@@ -1,4 +1,4 @@
-# Copyright (C) 2012 The Libphonenumber Authors
+# Copyright (C) 2016 The Libphonenumber Authors
 
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


resources/geocoding/en/98.txt
```diff
@@ -1,4 +1,4 @@
-# Copyright (C) 2012 The Libphonenumber Authors
+# Copyright (C) 2016 The Libphonenumber Authors
```
mizuni: So you can actually leave this as 2012 since the file was originally created then. We would only had the current calendar year for a new file.

resources/geocoding/en/98.txt
```diff
@@ -13,9 +13,7 @@
 # limitations under the License.
 
 # Generated from:
-# http://www.itu.int/oth/T0202000066/en [2012-07-24]
```
mizuni: I would recommend leaving all the old URLs for posterity's sake - especially the original ITU link, which is a stable URL where updates can usually (hopefully) be tracked consistently.

lararennie: Yes, I think the URL is fine, but the date should be changed.

Do we not change any prefixes now based on the other site?

behdad:
lgtm

behdad:
Combined into https://github.com/googlei18n/libphonenumber/pull/1185

PR #1157, opened 2016-06-11
Persian localized version of Afghanistan geocoding

afarsi:
I translated English Version of Afghanistan geocoding to Persian.
Note: One of two official languages in Afghanistan is Persian Dari Persian/ Dari Farsi.


behdad:
lgtm

PR #1170, opened 2016-06-21
Update checked-in jars by invoking mvn from root

keghani:
Diffs (see b/29343414):

~keghani/GitHub_diffs/e1ccea0/pkgdiff_reports/cpp-build/1.0-SNAPSHOT-jar-with-dependencies_to_1.0-SNAPSHOT-jar-with-dependencies_new/changes_report.html
~keghani/GitHub_diffs/e1ccea0/pkgdiff_reports/java-build/1.0-SNAPSHOT-jar-with-dependencies_to_1.0-SNAPSHOT-jar-with-dependencies_new/changes_report.html


java/pending_code_changes.txt
```diff
@@ -1 +1,5 @@
-
+Code changes:
+ - Added to checked-in tools/java/cpp-build and tools/java/java-build jars
```
lararennie: I am surprised that these depend on PhoneNumberUtil at all; it's a bit of a circular dependency, is it not? However I believe you - I remember having problems with this in the past. We should look at removing it...

Question though: do we want to update pending_code_changes.txt when we update these jars? they aren't really part of the project, they are just a tool to build the metadata. I don't think anyone is using these (or any other) methods through these jars: and indeed they shouldn't be..!

keghani:
Ready for another look.

lararennie:
So I patched your change, regenerated, and got diffs; presumably different java version used? What did you use?

keghani:
I would get diffs too - git does binary diffs between the jars, which could be affected by just how the files are ordered in the jar :(

Try pkgdiff? Or send me your jar (push your branch?) and I can pkgdiff.

PR #1171, opened 2016-06-21
Fix parent of libphonenumber-parent Maven project

keghani:
Project structure is unstable, <parent> by default refers to ../pom.xml unless specified otherwise - so java/pom.xml was using pom.xml as the parent, and had the wrong groupId and artifactId (b/29343414).

Tested pom.xml changes by running `mvn package` inside java/ - everything builds successfully.


keghani:
I'll send a new PR with the latest changes from master.

PR #1172, opened 2016-06-21
Test

lararennie:
Keghani; here are the jars when I run mvn package.


keghani:
Interesting, we seem to be running different protocs but I don't know why you're not running the generated one.
We're also using different pom.xmls?
And I don't understand why I don't have the PhoneNumberUtil updates, I thought I did in the last diff...

I'll have to investigate later, here are the results:

~keghani/GitHub_diffs/d755217/pkgdiff_reports/java-build/1.0-SNAPSHOT-jar-with-dependencies_keghani_to_1.0-SNAPSHOT-jar-with-dependencies_lara/changes_report.html
~keghani/GitHub_diffs/d755217/pkgdiff_reports/cpp-build/1.0-SNAPSHOT-jar-with-dependencies_keghani_to_1.0-SNAPSHOT-jar-with-dependencies_lara/changes_report.html

keghani:
This build should be hermetic as of #1282, we should get the same results.

PR #1192, opened 2016-07-05
Fix Maven project hierarchy and plugin versions

keghani:
Fixes b/22840154 and b/29343414.
Tested by running `mvn package` from root.

Still have javadoc warnings and errors (b/29977680, b/27241869, b/27241398), but everything else is cleaned up with this change.

Does not check in changed generated jars because their generation is currently not hermetic and outdated anyway (b/29546716); the output would not change either as a result of the changes in this PR in `tools/java/cpp-build/pom.xml` and `tools/java/java-build/pom.xml`, these changes actually bring us closer to being hermetic (b/24799921).

### DIFF
--- a/java/demo/pom.xml
+++ b/java/demo/pom.xml
@@ -31,7 +31,7 @@
       <version>2.5</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
+      <groupId>commons-io</groupId>
       <artifactId>commons-io</artifactId>
       <version>1.3.2</version>
     </dependency>
--- a/java/libphonenumber/pom.xml
+++ b/java/libphonenumber/pom.xml
@@ -31,6 +31,7 @@
     <plugins>
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
+        <version>2.2</version>
         <configuration>
           <excludes>
             <exclude>com/google/i18n/phonenumbers/SingleFileMetadataSourceImpl.class</exclude>
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -8,9 +8,9 @@
   <url>https://github.com/googlei18n/libphonenumber/</url>
 
   <parent>
-    <groupId>org.sonatype.oss</groupId>
-    <artifactId>oss-parent</artifactId>
-    <version>7</version>
+    <groupId>com.google.i18n.phonenumbers</groupId>
+    <artifactId>libphonenumber-build-parent</artifactId>
+    <version>1.0-SNAPSHOT</version>
   </parent>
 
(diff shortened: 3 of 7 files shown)


java/demo/pom.xml
```diff
@@ -31,7 +31,7 @@
       <version>2.5</version>
     </dependency>
     <dependency>
-      <groupId>org.apache.commons</groupId>
+      <groupId>commons-io</groupId>
```
keghani: Apparently this artifact moved.

java/libphonenumber/pom.xml
```diff
@@ -31,6 +31,7 @@
     <plugins>
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
+        <version>2.2</version>
```
keghani: These are just the version numbers my setup was using by default I suppose, I don't have a strong preference if you want different versions. This works.

java/pom.xml
```diff
@@ -8,9 +8,9 @@
   <url>https://github.com/googlei18n/libphonenumber/</url>
 
   <parent>
-    <groupId>org.sonatype.oss</groupId>
-    <artifactId>oss-parent</artifactId>
-    <version>7</version>
+    <groupId>com.google.i18n.phonenumbers</groupId>
```
keghani: More background on this change at https://github.com/googlei18n/libphonenumber/pull/1191

java/pom.xml
```diff
@@ -92,7 +92,7 @@
         <artifactId>maven-surefire-plugin</artifactId>
         <version>2.12</version>
         <configuration>
-          <forkMode>never</forkMode>
+          <forkCount>0</forkCount>
```
keghani: forkMode is deprecated, was giving some ambiguous warning about how this is incompatible with the default version of some other flag we're not even setting.

tools/java/data/pom.xml
```diff
@@ -47,16 +47,13 @@
...
         <groupId>org.apache.maven.plugins</groupId>
         <artifactId>maven-compiler-plugin</artifactId>
+        <version>2.0.2</version>
         <configuration>
           <source>1.6</source>
           <target>1.6</target>
         </configuration>
       </plugin>
       <plugin>
-        <groupId>net.kindleit</groupId>
```
keghani: Unused

tools/java/data/pom.xml
```diff
@@ -81,13 +78,6 @@
         </executions>
       </plugin>
       <plugin>
-        <groupId>org.apache.maven.plugins</groupId>
```
keghani: Unused

PR #1210, opened 2016-07-15
Silence "unstable build" warnings from Maven

keghani:
The reason I think this will work while the changes in https://github.com/googlei18n/libphonenumber/pull/1192 failed to deploy: this is the workaround at https://issues.apache.org/jira/browse/MNG-5146.

I'll release separately from metadata updates to e2e "test"/run this.
Fixes b/29343414.


keghani:
Also doesn't deploy...

PR #1213, opened 2016-07-15
Make MetadataSource implementations consistent

keghani:
Fixes b/24455712 along with cl/127311305.


keghani:
On hold until I change some things (as discussed offline with @hagbard).

keghani:
I'll send something else after refactorings.

PR #1206, opened 2016-07-14
Translating into Armenian

mhamlet:
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1206)

PR #1214, opened 2016-07-15
Orange to Ucom

armleo:


keghani:
Thanks for catching this @armleo! There is already a PR for this: https://github.com/googlei18n/libphonenumber/pull/1206, closing as duplicate.

PR #1254, opened 2016-08-09
Close ObjectInputStream after loading metadata

keghani:
Also made it clearer what's throwing the exceptions.

b/30726034


keghani:
I'll send something else.

PR #1256, opened 2016-08-09
Refactor metadata loading and close all streams after loading metadata

keghani:
b/30669304
b/30726034
b/30807096

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
@@ -66,7 +66,7 @@ class MetadataManager {
       try {
         in.close();
       } catch (IOException e) {
-        LOGGER.log(Level.WARNING, e.toString());
+        LOGGER.log(Level.WARNING, "error closing input stream (ignored)", e);
       }
     }
   }
@@ -89,22 +89,40 @@ class MetadataManager {
     return CodedInputByteBufferNano.newInstance(outputStream.toByteArray());
   }
 
-  private static void loadAlternateFormatsMetadataFromFile(int countryCallingCode) {
-    InputStream source = PhoneNumberMatcher.class.getResourceAsStream(
-        ALTERNATE_FORMATS_FILE_PREFIX + "_" + countryCallingCode);
-    ObjectInputStream in = null;
+  /**
+   * Loads and returns the metadata protocol buffer from the given stream and closes the stream.
+   */
+  static PhoneMetadataCollection loadMetadataAndCloseInput(InputStream source, int bufferSize) {
+    ObjectInputStream ois;
     try {
-      in = new ObjectInputStream(source);
-      CodedInputByteBufferNano byteBuffer = convertStreamToByteBuffer(in, BUFFER_SIZE);
-      PhoneMetadataCollection alternateFormats = new PhoneMetadataCollection();
-      alternateFormats.mergeFrom(byteBuffer);
-      for (PhoneMetadata metadata : alternateFormats.metadata) {
-        callingCodeToAlternateFormatsMap.put(metadata.countryCode, metadata);
-      }
+      ois = new ObjectInputStream(source);
     } catch (IOException e) {
-      LOGGER.log(Level.WARNING, e.toString());
+      close(source);
+      throw new RuntimeException("cannot load/parse metadata", e);
+    }
+
+    try {
+      PhoneMetadataCollection metadataCollection = new PhoneMetadataCollection();
+      try {
+        metadataCollection.mergeFrom(convertStreamToByteBuffer(ois, bufferSize));
+      } catch (IOException e) {
+        throw new RuntimeException("cannot load/parse metadata", e);
+      }
+      return metadataCollection;
     } finally {
-      close(in);
+      close(ois);
+    }
+  }
+
+  private static void loadAlternateFormatsMetadataFromFile(int countryCallingCode) {
+    String fileName = ALTERNATE_FORMATS_FILE_PREFIX + "_" + countryCallingCode;
+    InputStream source = PhoneNumberMatcher.class.getResourceAsStream(fileName);
+    if (source == null) {
+      throw new IllegalStateException("missing metadata: " + fileName);
+    }
+    PhoneMetadataCollection metadataCollection = loadMetadataAndCloseInput(source, BUFFER_SIZE);
+    for (PhoneMetadata metadata : metadataCollection.metadata) {
+      callingCodeToAlternateFormatsMap.put(metadata.countryCode, metadata);
     }
   }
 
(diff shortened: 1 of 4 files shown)


keghani:
cc @hagbard

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -89,22 +89,40 @@ static CodedInputByteBufferNano convertStreamToByteBuffer(ObjectInputStream in,
...
-      CodedInputByteBufferNano byteBuffer = convertStreamToByteBuffer(in, BUFFER_SIZE);
-      PhoneMetadataCollection alternateFormats = new PhoneMetadataCollection();
-      alternateFormats.mergeFrom(byteBuffer);
-      for (PhoneMetadata metadata : alternateFormats.metadata) {
-        callingCodeToAlternateFormatsMap.put(metadata.countryCode, metadata);
-      }
+      ois = new ObjectInputStream(source);
     } catch (IOException e) {
-      LOGGER.log(Level.WARNING, e.toString());
+      close(source);
```
lararennie: So it seems source is only closed if there is an IOException? Should it not be closed in a finally block?

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -121,21 +139,14 @@ static PhoneMetadata getAlternateFormatsForCountry(int countryCallingCode) {
...
-    try {
-      in = new ObjectInputStream(source);
-      CodedInputByteBufferNano byteBuffer = convertStreamToByteBuffer(in, BUFFER_SIZE);
-      PhoneMetadataCollection shortNumberMetadata = new PhoneMetadataCollection();
-      shortNumberMetadata.mergeFrom(byteBuffer);
-      for (PhoneMetadata metadata : shortNumberMetadata.metadata) {
+    String fileName = SHORT_NUMBER_METADATA_FILE_PREFIX + "_" + regionCode;
+    InputStream source = PhoneNumberMatcher.class.getResourceAsStream(fileName);
+    if (source == null) {
+      throw new IllegalStateException("missing metadata: " + fileName);
```
lararennie: Add a comment about why this should never happen (that we only load things based on the expectation that they are present, by checking the map of available data first).

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -121,21 +139,14 @@ static PhoneMetadata getAlternateFormatsForCountry(int countryCallingCode) {
...
-        SHORT_NUMBER_METADATA_FILE_PREFIX + "_" + regionCode);
-    ObjectInputStream in = null;
-    try {
-      in = new ObjectInputStream(source);
-      CodedInputByteBufferNano byteBuffer = convertStreamToByteBuffer(in, BUFFER_SIZE);
-      PhoneMetadataCollection shortNumberMetadata = new PhoneMetadataCollection();
-      shortNumberMetadata.mergeFrom(byteBuffer);
-      for (PhoneMetadata metadata : shortNumberMetadata.metadata) {
+    String fileName = SHORT_NUMBER_METADATA_FILE_PREFIX + "_" + regionCode;
+    InputStream source = PhoneNumberMatcher.class.getResourceAsStream(fileName);
```
lararennie: PhoneNumberMatcher? I am confused that the PhoneNumberMatcher class has the short number metadata at all?

java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java
```diff
@@ -110,13 +110,18 @@ private boolean isNonGeographical(int countryCallingCode) {
   static <T> PhoneMetadata loadMetadataFromFile(
       T key, ConcurrentHashMap<T, PhoneMetadata> map, String filePrefix,
       MetadataLoader metadataLoader) {
+    // The size of the byte buffer used for deserializing the phone number metadata files for each
+    // region.
```
lararennie: You could define it at class level as a private final int constant.

Also do you know why it is this size, 16 *1024?

java/libphonenumber/src/com/google/i18n/phonenumbers/SingleFileMetadataSourceImpl.java
```diff
@@ -95,60 +95,29 @@ public PhoneMetadata getMetadataForNonGeographicalRegion(int countryCallingCode)
 
   // @VisibleForTesting
   void loadMetadataFromFile() {
+    // The size of the byte buffer for deserializing the single nano metadata file which holds
+    // metadata for all regions.
+    final int singleFileBufferSize = 256 * 1024;
```
lararennie: same here re making it a constant

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
keghani: Yes! But not here, if all succeeds we need the InputStream open to continue. Re-organized the try-blocks.

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
keghani: Done. Except in the single file impl, here we expect it to happen since the Maven jar doesn't package the single file :) No one should be using that unless they're packaging their own build.

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
keghani: This works when I change to MetadataManager! Done

java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java
keghani: Defined in MetadataManager, added comments (long-term I'll move all this logic there, doesn't make sense to have 3 extra files just for the regular loading).

java/libphonenumber/src/com/google/i18n/phonenumbers/SingleFileMetadataSourceImpl.java
keghani: Defined in MetadataManager, added comments

keghani:
PTAL

keghani:
Also, do you think this warrants a minor major version bump (x.y.0)? Although it fixes a rare bug it's not going to change how clients use the library at all, so I'd think not.

I think yes though for inclusion in release notes as code changes?

lararennie:
Code changes: yes mention. Minor/major version bump: no.

2016-08-16 15:10 GMT+02:00 Keghani Kouzoujian <notifications@github.com>:

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -61,19 +61,15 @@
...
-      try {
-        in.close();
-      } catch (IOException e) {
-        LOGGER.log(Level.WARNING, e.toString());
-      }
-    }
-  }
+  // The size of the byte buffer in bytes used to convert a stream containing metadata for a single
+  // region, to a nanoproto-compatible CodedInputByteBufferNano. This was determined by the size of
+  // the metadata with which we build our actual jars.
```
lararennie: How do you mean the size of the metadata? what does the 16 represent, what does the 1024 represent? Similarly below: why 256? It looks like a calculation but I'm confused what it is calculating.

java/libphonenumber/src/com/google/i18n/phonenumbers/SingleFileMetadataSourceImpl.java
```diff
@@ -97,58 +94,25 @@ public PhoneMetadata getMetadataForNonGeographicalRegion(int countryCallingCode)
   void loadMetadataFromFile() {
     InputStream source = metadataLoader.loadMetadata(fileName);
     if (source == null) {
-      logger.log(Level.SEVERE, "missing metadata: " + fileName);
+      // This should not happen since clients shouldn't be using this implementation!
```
lararennie: I don't understand this comment. If they shouldn't be using it, why is the code here?

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -89,22 +85,54 @@ static CodedInputByteBufferNano convertStreamToByteBuffer(ObjectInputStream in,
...
-    } catch (IOException e) {
-      LOGGER.log(Level.WARNING, e.toString());
+      return metadataCollection;
     } finally {
-      close(in);
+      try {
+        if (ois != null) {
+          ois.close();
+        } else {
+          source.close();
```
lararennie: why do we only close source() if ois != null? don't we promise always to do it?

keghani: ois.close() also closes source. But if ois is null (as in if the ObjectInputStream initialization failed) then we want to attempt closing source independenly.

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
keghani: I believe @dlyongemallo wrote the numbers this way to signify that 1024 bytes is a Kb, so this is 16 Kb (power-of-2 ceiling of the largest per-region binary metadata file with room for growth), and the other is 256 Kb (power-of-2 ceiling of the single binary metadata file containing all regions' metadata, with room for growth - we don't support clients using this - it's experimental - and we don't ship this file to Maven Central).

lararennie:
LGTM, thanks for the explanations :)

PR #1277, opened 2016-08-19
IsNumberGeographical public + better detection of "geographical numbers" in geocoder

lararennie:
Making function IsNumberGeographical public. This function operates on
the type of number and the country it belongs to only, so may have some
false positives.

Using this in the geocoder so that geocoding now limited to numbers that
we consider geographical, based on their type and country, rather than
just based on their type. The C++ geocoder did not previously check the
number type/country at all.

Indonesian and Chinese mobile numbers have now been added to the list of
possibly-geographical numbers.


keghani:
Huh. Travis is doing checks although we didn't merge that commit and this is a different branch without that commit.

lararennie:
Yes: it is interesting to me: I did configure it to, but it said it needed
the .yaml file: but maybe if it's absent, it assumes the one liner file I
added anyway :)

2016-08-19 10:51 GMT+02:00 Keghani Kouzoujian <notifications@github.com>:

lararennie:
I can revert the generated jars from this CL too if desired, let me know :) they are just there at the moment because it was easier. They are much smaller than they were before!!! I guess all your work removing stuff :)

cpp/src/phonenumbers/phonenumberutil.cc
```diff
@@ -386,9 +386,15 @@ class PhoneNumberRegExpsAndMappings {
 
     mobile_token_mappings_.insert(std::make_pair(52, '1'));
     mobile_token_mappings_.insert(std::make_pair(54, '9'));
+    geo_mobile_countries_without_mobile_area_codes_.insert(86);  // Chin
```
keghani: s/Chin/China :)

cpp/src/phonenumbers/phonenumberutil.h
```diff
@@ -489,6 +489,20 @@ class PhoneNumberUtil : public Singleton<PhoneNumberUtil> {
       const string& number,
       const string& region_dialing_from) const;
 
+  // Tests whether a phone number has a geographical association. It checks if
+  // the number is associated to a certain region in the country where it
+  // belongs to. Note that this doesn't verify if the number is actually in use.
+  bool IsNumberGeographical(const PhoneNumber& phone_number) const;
```
keghani: These 2 should come right after void TrimUnwantedEndChars(string* number) const;
?

cpp/src/phonenumbers/phonenumberutil.cc
lararennie: Done :)

cpp/src/phonenumbers/phonenumberutil.h
lararennie: No, because they have to be public now so the geocoder can use them. If the geocoder didn't exist they could be private.

cpp/test/phonenumbers/phonenumberutil_test.cc
```diff
@@ -1341,7 +1341,8 @@ TEST_F(PhoneNumberUtilTest, GetLengthOfGeographicalAreaCode) {
   number.set_national_number(2070313000ULL);
   EXPECT_EQ(2, phone_util_.GetLengthOfGeographicalAreaCode(number));
 
-  // A UK mobile phone, which has no area code.
+  // A mobile number in the UK does not have an area code (by default, mobile
+  // numbers do not, unless they have been added to our list of exceptions).
   number.set_country_code(44);
   number.set_national_number(7123456789ULL);
```
keghani: Hmm, this is different from in [PhoneNumberUtilTest.java](https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberUtilTest.java). There GB_MOBILE has 7912345678 for the national number.

Also shouldn't PhoneNumberUtilTest.java be updated with these changes?

cpp/src/phonenumbers/phonenumberutil.h
keghani: I see, thanks!

cpp/test/phonenumbers/phonenumberutil_test.cc
```diff
@@ -1444,6 +1450,12 @@ TEST_F(PhoneNumberUtilTest, GetLengthOfNationalDestinationCode) {
   number.set_country_code(800);
   number.set_national_number(12345678ULL);
   EXPECT_EQ(4, phone_util_.GetLengthOfNationalDestinationCode(number));
+
+  // A mobile number from China is geographical, but does not have an area code.
```
keghani: Move this test to the test method testing GetLengthOfGeographicalAreaCode :)

And add a test for GetLengthOfNationalDestinationCode for China with comment:

```
  // A mobile number from China is geographical, but does not have an area code:
  // however it still can be considered to have a national destination code.
```

keghani:
Please also update https://github.com/googlei18n/libphonenumber/blob/master/resources/PhoneNumberMetadataForTesting.xml

Also rebase on top of master (to pick up the smart quotes change) then regenerate the jar files. Then I'll double-check those and should be good to merge!

lararennie:
Added xml file. Will rebase :)

cpp/test/phonenumbers/phonenumberutil_test.cc
lararennie: Yes and yes done :)

cpp/test/phonenumbers/phonenumberutil_test.cc
lararennie: Think I've done this :)

cpp/test/phonenumbers/phonenumberutil_test.cc
```diff
@@ -1444,6 +1456,13 @@ TEST_F(PhoneNumberUtilTest, GetLengthOfNationalDestinationCode) {
   number.set_country_code(800);
   number.set_national_number(12345678ULL);
   EXPECT_EQ(4, phone_util_.GetLengthOfNationalDestinationCode(number));
+
+  // A mobile number from China is geographical, but does not have an area code:
+  // however it still can be considered to have a national destination code.
+  PhoneNumber cn_mobile;
+  cn_mobile.set_country_code(86);
+  cn_mobile.set_national_number(18912341234ULL);
+  EXPECT_EQ(0, phone_util_.GetLengthOfGeographicalAreaCode(cn_mobile));
```
keghani: This is still calling GetLengthOfGeographicalAreaCode, should change to GetLengthOfNationalDestinationCode?

keghani:
PTAL - one comment + regenerate jars if you care, otherwise we can submit as is and regenerate later.

cpp/test/phonenumbers/phonenumberutil_test.cc
lararennie: Good catch...!

keghani:
b/30355729

PR #1272, opened 2016-08-18
Enabling continuous build for Java with Travis.

lararennie:

Issue #1070, opened 2016-04-07
javanano is not available in protobuf 2.6.1


puntogil:
Hi
any ideas?

[INFO] Skipping libphonenumber-build-parent
[INFO] This project has been banned from the build due to previous failures.
[INFO] ------------------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] Libphonenumber build tools ......................... SUCCESS [  0.005 s]
[INFO] Libphonenumber common library for build tools ...... FAILURE [  4.440 s]
[INFO] Libphonenumber C++ build tools ..................... SKIPPED
[INFO] Libphonenumber Data tools .......................... SKIPPED
[INFO] Libphonenumber Java and JavaScript build tools ..... SKIPPED
[INFO] libphonenumber-parent .............................. SKIPPED
[INFO] libphonenumber ..................................... SKIPPED
[INFO] prefixmapper ....................................... SKIPPED
[INFO] carrier ............................................ SKIPPED
[INFO] geocoder ........................................... SKIPPED
[INFO] libphonenumber-build-parent ........................ SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 9.793 s
[INFO] Finished at: 2016-04-07T11:04:38+02:00
[INFO] Final Memory: 21M/210M
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.3:compile (default-compile) on project common-build: Compilation failure: Compilation failure:
[ERROR] ~/rpmbuild/BUILD/libphonenumber-libphonenumber-7.3.0/java/libphonenumber/src/com/google/i18n/phonenumbers/nano/Phonemetadata.java:[188,31] package com.google.protobuf.nano does not exist
[ERROR] ~/rpmbuild/BUILD/libphonenumber-libphonenumber-7.3.0/java/libphonenumber/src/com/google/i18n/phonenumbers/nano/Phonemetadata.java:[226,49] package com.google.protobuf.nano does not exist
[ERROR] ~/rpmbuild/BUILD/libphonenumber-libphonenumber-7.3.0/java/libphonenumber/src/com/google/i18n/phonenumbers/nano/Phonemetadata.java:[260,37] package com.google.protobuf.nano does not exist
[...]

Issue #956, opened 2016-01-27
Building the Java package with debuild fails at HEAD


roubert:
Building the libphonenumber Debian packages directly from a GitHub checkout does no longer work, but fails while building the Java packages:

`$ debuild -i -us -uc -b`

`[ERROR] /usr/local/google/home/roubert/src/libphonenumber/googlei18n/java/libphonenumber/src/com/google/i18n/phonenumbers/nano/Phonemetadata.java:[302,30] error: package com.google.protobuf.nano does not exist`

(See attached log file for all details.)
[debuild.log.txt](https://github.com/googlei18n/libphonenumber/files/107266/debuild.log.txt)

roubert:
@dpocock I think you were the last one to make changes here. Can it be as easy as a new dependency that needs to be listed somewhere?

keghani:
@roubert 
We moved to nano/Phonemetadata.java and added dependencies (almost everywhere apparently) at release 7.2.1 - likely we missed this.

keghani:
In logs: found protobuf, version 2.5.0

2.5.0 doesn't have nanoproto - to avoid your having to install the required versions we did this hermetically with Maven: https://github.com/googlei18n/libphonenumber/commit/d8268984e4333a6a9215310b9c0c2f376b4cad85

Perhaps something similar can be done with the Debian build.

TimurFayruzov:
I had the same issue when building libphonenumber for java locally and it turns out I had to install the protobuf library in local maven repository, which resolved the problem:
```
mvn org.apache.maven.plugins:maven-install-plugin:2.5.2:install-file -Dfile=../lib/com/google/protobuf/protobuf-java/2.6.1/protobuf-java-2.6.1.jar -DpomFile=../lib/com/google/protobuf/protobuf-java/2.6.1/protobuf-java-2.6.1.pom
```

keghani:
Fixed in 7.6.0 I think?
It would be great if someone could verify :)

See release notes, you may need to add a dependency (but this time it will be standard): https://github.com/googlei18n/libphonenumber/releases/tag/libphonenumber-7.6.0.

PR #1282, opened 2016-08-19
Maven to use protobuf-javanano 3.0.0-alpha-7 from Maven Central

### LINKED ISSUE #913: Make Protobuf dependency optional
Since 7.2.0 `protobuf-java` is included as a dependency (as a jar in 7.2.1). On Android this can be an issue as it increases the dex method count for this library from 660 to **7106**.

The total dex method limit for an Android APK is ~65k, so this increase makes it very difficult to upgrade.

Using gradle I'm able to exclude the compile time protobuf dependency in 7.2.0, but since 7.2.1 it is included as a jar and I'm not sure how to exclude it.

    compile ('com.googlecode.libphonenumber:libphonenumber:7.2.0') {
        exclude module: 'protobuf-java'
    }

Is it possible to exclude this dependency when it is included as a jar?
If not, can it be changed back to a compile time dependency?

keghani:
After this commit, builds and tests succeeded and generated files didn't change (except for metadata changes to master since the last release, plus minor Maven point, see below**) when invoking:

* `ant junit` from `java/`
* Ant targets for metadata and test generation as during release
* `mvn package` from root
* Travis continuous integration on top of this commit (Thanks @lararennie *may not insert emoji in markdown but :)*): #1283

See readme rendered [here](https://github.com/googlei18n/libphonenumber/pull/1282/files?short_path=6721f0c#diff-6721f0c53f24c6a08eedb158c4977e63).

Jar diffs of this commit are due to protocol buffer version differences.

**If I run `mvn package` at the end I get `java-build` and `cpp-build` jars that trigger git diffs (and would be tracked as changes in a commit) because the binary files differ; but if I unzip the old and new jars then `diff -r` them, their contents are identical.

Fixes
#913 Make Protobuf dependency optional
#956 Building the Java package with debuild fails at HEAD
#1070 javanano is not available in protobuf 2.6.1

Unblocks
#1272 Enabling continuous build for Java with Travis.


lararennie:
LGTM :)

PR #1290, opened 2016-08-24
Maven updates 7.6.0

keghani:


keghani:
<using a different branch - branched off of master with the release changes - since the original diverged after squashed commit>

PR #1292, opened 2016-08-25
Change Maven parent of libphonenumber-parent artifact

keghani:
This change results in a build success for these calls from the root directory:

```
ant -f java/build.xml \
    build-phone-metadata build-test-metadata \
    build-short-metadata build-alternate-metadata \
    build-timezones-data build-timezones-test-data \
    junit
mvn package
mvn install
```

Yet when we got to deployment last time, we failed. Sending a PR to see whether Travis would catch this.

### DIFF
--- a/java/pom.xml
+++ b/java/pom.xml
@@ -8,9 +8,9 @@
   <url>https://github.com/googlei18n/libphonenumber/</url>
 
   <parent>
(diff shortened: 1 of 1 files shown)


keghani:
Conclusion: Travis can't help us determine whether it will deploy.

PR #1296, opened 2016-08-26
Use protobuf-lite

azzar1:
Use libprotobuf-lite in cpp/

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -100,9 +100,9 @@ if (${USE_RE2} STREQUAL "ON")
   find_required_library (RE2 re2/re2.h re2 "Google RE2")
 endif ()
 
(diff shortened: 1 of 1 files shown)


azzar1:
I signed it!

keghani:
Hi @azzar1! What issue does this fix? I can build C++ with the current setup.

azzar1:
Hi @keghani, for what I see libphonenumber does not need to be build with the full libprotobuf. the lite version seems to be enough. So I was wondering why libphonenumber still links to the full version.

keghani:
Mostly because there's no issue with this as far as I'm aware. Linking against libprotobuf-lite would require us to re-test everything and update the [README](https://github.com/googlei18n/libphonenumber/blob/master/cpp/README), and it's hard to invest in this now considering nothing's broken.

Issue #205, opened 2014-12-03
OSGI Buncle packaging


mihaelacr-google:
Imported from [Google Code issue #205](https://code.google.com/p/libphonenumber/issues/detail?id=205) created by [reda.bendiar](https://code.google.com/u/112954534242903792131/) on 2012-10-17T17:03:31.000Z:

----
Is it possible to have the java version packaged as osgi bundle ?

keghani:
I don't know whether a CLA was signed for that patch, or whether it's enough for the process to be automatic along with our regular releases.
I'll close this issue as there hasn't been interest or contributions since we moved to GitHub.
Please re-open if you can help!

frett:
I'm looking to utilize this library in an OSGi environment, so I resurrected the patch and made a couple minor adjustments

PR #1300, opened 2016-08-29
add OSGi meta-data to libphonenumber and geocoder

frett:
This takes the patch submitted for #205 and modifies it to be a bit less intrusive with how it adds the OSGi meta-data.

This patch follows the mechanism described here: http://felix.apache.org/documentation/subprojects/apache-felix-maven-bundle-plugin-bnd.html#adding-osgi-metadata-to-existing-projects-without-changing-the-packaging-type

How this works is that the `maven-bundle-plugin` is used to generate a `MANIFEST.MF` containing the necessary OSGi meta-data at build time, then the `maven-jar-plugin` is configured to utilize this built `MANIFEST.MF` file instead of generating it's own, everything else about the build process is left untouched.

### DIFF
--- a/java/geocoder/pom.xml
+++ b/java/geocoder/pom.xml
@@ -38,26 +38,6 @@
     </testResources>
     <plugins>
       <plugin>
-        <groupId>org.apache.felix</groupId>
-        <artifactId>maven-bundle-plugin</artifactId>
-        <version>3.2.0</version>
-        <configuration>
-          <instructions>
-            <Export-Package>com.google.i18n.phonenumbers.geocoding</Export-Package>
-            <Private-Package>com.google.i18n.phonenumbers.geocoding.data</Private-Package>
-          </instructions>
-        </configuration>
-      </plugin>
-      <plugin>
-        <artifactId>maven-jar-plugin</artifactId>
-        <version>3.0.2</version>
-        <configuration>
-          <archive>
-            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-          </archive>
-        </configuration>
-      </plugin>
-      <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>animal-sniffer-maven-plugin</artifactId>
         <version>1.15</version>
--- a/java/geocoder/pom.xml
+++ b/java/geocoder/pom.xml
@@ -38,6 +38,24 @@
     </testResources>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <Export-Package>com.google.i18n.phonenumbers.geocoding</Export-Package>
+            <Private-Package>com.google.i18n.phonenumbers.geocoding.data</Private-Package>
+          </instructions>
+        </configuration>
+      </plugin>
+      <plugin>
+        <artifactId>maven-jar-plugin</artifactId>
+        <configuration>
+          <archive>
+            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+          </archive>
+        </configuration>
+      </plugin>
+      <plugin>
         <groupId>org.codehaus.mojo</groupId>
         <artifactId>animal-sniffer-maven-plugin</artifactId>
         <version>1.15</version>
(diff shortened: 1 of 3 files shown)


frett:
signed the CLA

java/libphonenumber/pom.xml
```diff
@@ -30,6 +30,24 @@
     </testResources>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
```
keghani: Please add version number here and other plugins as well to keep the Maven build hermetic.

java/geocoder/pom.xml
```diff
@@ -38,6 +38,26 @@
     </testResources>
     <plugins>
       <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <version>3.2.0</version>
+        <configuration>
+          <instructions>
+            <Export-Package>com.google.i18n.phonenumbers.geocoding</Export-Package>
```
lararennie: This excludes the PhoneNumberToTimeZonesMapper: is this intentional?

java/libphonenumber/pom.xml
```diff
@@ -30,6 +30,26 @@
...
+        <version>3.2.0</version>
+        <configuration>
+          <instructions>
+            <Export-Package>com.google.i18n.phonenumbers</Export-Package>
+            <Private-Package>com.google.i18n.phonenumbers.data</Private-Package>
```
lararennie: We only have resource files in the data folder; same above - is this intended to cover them? but they don't have a "package" per se?

lararennie:
In general happy with the idea, although we don't commit to having this always working if it becomes to difficult in future.

How have you tested this?

frett:
I have tested out the libphonenumber library with the OSGi meta-data. This is the one we are using in our project that requires OSGi. We use parsing, validation, and formatting functionalities from it and the OSGi meta-data covers those use cases without any issues.

I have not tested out the geocoder library, I included the OSGi meta-data for it because it was in the original patch that I was resurrecting.

frett:
Of note, I also had to make a minor adjustment to the OSGi meta-data included in the javanano protobuf library. PR for that was here: https://github.com/google/protobuf/pull/2033

java/libphonenumber/pom.xml
frett: Based on my understanding of OSGi, these directives only have an impact on how Classes are exposed in an OSGi environment. Since there are only resources within the data directory this statement would actually have no impact on the generated meta-data. I'm going to remove it since it has no actual effect.

java/geocoder/pom.xml
frett: I hadn't tested out the geocoder module, I just took the rules that were defined in the previous patch and included them.

Unfortunately, I believe `PhoneNumberToTimeZonesMapper` being in the same package as all of libphonenumber will cause wiring issues within an OSGi environment if it's exported.

Within an OSGi environment each jar/bundle gets it's own ClassLoader. For external imports the environment links these ClassLoaders together based on the Export-Package meta-data. The environment chooses a single ClassLoader per imported package (not class) to link. This would mean that exporting `com.google.i18n.phonenumbers` within the geocoder jar would cause a conflict with the main libphonenumber jar.

frett:
Actually looking closer at the geocoder OSGi metadata and actual java code, it wouldn't yield a jar that would actually be usable for anything within an OSGi environment.

The module provides 2 classes currently.

1. `PhoneNumberToTimeZonesMapper`, which can't be exported due to a package conflict with libphonenumber within OSGi environments.
2. `PhoneNumberOfflineGeocoder`, which has a dependency on the prefixmapper module which doesn't include it's own OSGi metadata, plus has potential resource ClassLoader issues within OSGi (#577)

So, neither of these classes would be usable in OSGi in the current state. My personal uses don't require geocoder, so I'm fine dropping the geocoder OSGi metadata. Or, alternatively I could add meta-data to prefixmapper, but that would also require that cross-jar resource loading is addressed for OSGi environments (#577).

I would lean towards dropping the metadata, but I'll let it be your call as to how you want to proceed with geocoder.

lararennie:
I think dropping geocoding for now is sensible. Best to have something that works for the average use-case.

lararennie:
Thanks for this! I won't merge today because we might be doing a new release that we want to only have one change in it; if we don't do that, I'll merge; or if we do, I'll merge this after that happens (sometime next week).

PR #1312, opened 2016-09-06
Add hash as a diallable char.

tiagosh:
This is a patch we have in the libphonenumber package on ubuntu for quite some time now and would be nice to get it upstream.

According to ofono docs, the hash is a diallable char: http://git.kernel.org/cgit/network/ofono/ofono.git/tree/doc/voicecallmanager-api.txt#n41

---

This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1312)


lararennie:
One issue we have is that # is frequently written in phone numbers *not* for dialling, but to indicate an extension suffix, e.g. 1-800 543-1234#677 -> doesn't mean dial the#, it means 677 is an extension. Or there is the variant with that at the end. Also people provide phone numbers like "PH#886-2-25450000" and this # is also not for dialling, it is just saying "number". So in that situation, actually the # should be stripped....

Not saying we shouldn't change this - we should discuss, but if we do, we need some docs around this.

What do you guys use this method for?

tiagosh:
Hi, thanks for the quick reply.

I believe we added this patch to cover this specific use case: https://bugs.launchpad.net/ubuntu/+source/dialer-app/+bug/1366331

We use libphonenumber to format numbers in the dialer application as the user types, and I personally always thought # was not allowed for dialing, but as per reporter comments, this use case isn't USSD (as the number does not start with a * or #) and also as per ofono doc, the Dial() method description says # is a valid char, so I concluded I was wrong and then created the patch.

lararennie:
I'm inclined to accept this; we already accept * which we don't store in the phone number object either, plus you're right, it's according to standard. Also it happens after extension parsing so we've already extracted the # in extensions fine. I'll do some tests to see if there are any surprising side-effects and get back to you.

lararennie:
Right, back on this :)
Can you please update:
https://github.com/googlei18n/libphonenumber/blob/master/java/pending_code_changes.txt with a short description of the change
And:
PhoneNumberUtil.java: with the same change (and in the unit-test)
And:
https://github.com/googlei18n/libphonenumber/blob/master/javascript/i18n/phonenumbers/phonenumberutil.js
- although the method is not exposed there, not sure if there are any unit-tests that need updating, probably not.

lararennie:
Hi Tiago,

Are you able to finish this patch? Or do I need to take over and do it instead?

Thanks!

tiagosh:
Hello,

Just updated the branch as you requested.
Let me know if any other change is required.
Thanks

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberUtilTest.java
```diff
@@ -427,7 +427,7 @@ public void testNormaliseStripAlphaCharacters() {
 
   public void testNormaliseStripNonDiallableCharacters() {
     String inputNumber = "03*4-56&+a#234";
-    String expectedOutput = "03*456+234";
+    String expectedOutput = "03*456+#234";
```
lararennie: This should be the same as the C++ test, where you added a 1 in front of the a as well in the inputNumber.

lararennie:
Two last small comments, otherwise LGTM, thanks!!

lararennie:
OK
Thanks so much for the changes and the contribution!

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/googlei18n/libphonenumber/1312#-:-KVZKJFMx9n5GDBEcqjF:bqwrnuh)*

lararennie:
:lgtm:

---

Review status: 0 of 6 files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/googlei18n/libphonenumber/1312#-:-KVZKQPjEHGzZSERiDkj:bnfp4nl)*

lararennie:
Reviewed 1 of 2 files at r3, 2 of 4 files at r4, 1 of 2 files at r5, 2 of 2 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/googlei18n/libphonenumber/1312)*

PR #1339, opened 2016-09-15
Revert use of protobuf-javanano (part 1)

keghani:
We remove the dependency on protobuf (protobuf-javanano) from the Java library.
Resurrecting java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java which was present until libphonenumber-7.1.1.
Does not affect the C++ or JavaScript libraries.

** Begin note to reviewers, do not copy into final commit **

I will follow up with the generated file changes, either in this PR or separately into the proposed (non-master) branch, after which all can be merged to master.

For now this is all the hand-written changes, please review.
Proof of concept that once other files are generated this will build end-to-end: https://github.com/googlei18n/libphonenumber/pull/1338

Interesting parts to review:

- java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java:
This is almost what it was in 7.1.1; I added hand-crafted API for possible lengths.
You can diff https://raw.githubusercontent.com/keghani/libphonenumber/7.7.0/java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java and https://raw.githubusercontent.com/googlei18n/libphonenumber/libphonenumber-7.1.1/java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java locally.

- tools/java/common/src/com/google/i18n/phonenumbers/BuildMetadataFromXml.java

- tools/java/common/test/com/google/i18n/phonenumbers/BuildMetadataFromXmlTest.java

- tools/java/cpp-build/pom.xml

- tools/java/cpp-build/src/com/google/i18n/phonenumbers/BuildMetadataCppFromXml.java is back to what it was in 7.1.1.

- tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataJsonFromXml.java too; interestingly the nanoproto version had a slightly different logic where it was writing the boolean nationalPrefixOptionalWhenFormatting (as true) when it was true (otherwise writing null), but before (and now) we're writing it (as true or false) when it's set (hasNationalPrefixOptionalWhenFormatting())

- When I first built the entire PR I generated C++ metadata. But now I don't generate anything, even when I try from the same place (in that branch, resetting head to the previous commit). Unsure what's going on. Originally I had this: https://github.com/keghani/libphonenumber/commit/b510d075c6d7852f54fdc135d9fe286a2a92a0f4

** End note to reviewers **


keghani:
@lararennie @andyst please review :)

.travis.yml
```diff
@@ -1,5 +1,7 @@
 language: java
+install:
+- mvn install -P travis -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
lararennie: What do these lines do?

tools/java/cpp-build/pom.xml
```diff
@@ -44,16 +64,42 @@
             </goals>
             <configuration>
               <sources>
+                <!-- Add generated/ to the source directories. -->
+                <!-- TODO: Add needed functionality to java/libphonenumber/src/Phonemetadata.java
```
lararennie: I wouldn't say necessarily "add needed functionality" because that doesn't solve the problem of it potentially getting out of sync.. Also depending on a maven release might be fine if we specified the version and said the C++ one had to be the same. I would state the problem and leave it open in the TODO how we are going to solve it.

tools/java/pom.xml
```diff
@@ -19,11 +19,27 @@
...
+        <module>common</module>
+        <module>cpp-build</module>
+        <module>data</module>
+        <module>java-build</module>
+      </modules>
+    </profile>
+    <profile>
+      <id>travis</id>
+      <modules>
+        <module>common</module>
```
lararennie: Comment here why the cpp-build is missing from this?

.travis.yml
keghani: The `-P travis` ensures that the `travis` profile is invoked from the `pom.xml` file [1][2], which I configured not to build the `cpp-build` module.
This command will run by default during the "install" phase by Travis, without the `-P travis` though by default [3][4]. I explicitly added it in this file to add the argument.

Do you think I should add a comment in this file? I think I can do that with a `#`.

[1] http://stackoverflow.com/questions/8304110/skip-a-submodule-during-a-maven-build
[2] http://maven.apache.org/guides/introduction/introduction-to-profiles.html
[3] https://docs.travis-ci.com/user/languages/java#Dependency-Management
[4] https://docs.travis-ci.com/user/customizing-the-build#The-Build-Lifecycle

tools/java/cpp-build/pom.xml
keghani: Actually I made a mistake about the Maven version - I had thought that it was getting protoc from the protobuf-java dependency but it's not (that's used for the runtime dependencies, not to generate the files).

Changed the comments, I'll push the changes momentarily.

tools/java/pom.xml
keghani: Done (commit coming up)

keghani:
Thanks! Address first round of comments.

.travis.yml
lararennie: Yes, if it works :)

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -130,10 +104,9 @@ private static void loadAlternateFormatsMetadataFromFile(int countryCallingCode)
       // that they are present, by checking the map of available data first.
       throw new IllegalStateException("missing metadata: " + fileName);
     }
-    PhoneMetadataCollection alternateFormatData =
-        loadMetadataAndCloseInput(source, DEFAULT_BUFFER_SIZE);
-    for (PhoneMetadata metadata : alternateFormatData.metadata) {
-      callingCodeToAlternateFormatsMap.put(metadata.countryCode, metadata);
+    PhoneMetadataCollection alternateFormats = loadMetadataAndCloseInput(source);
```
lararennie: can we change this back to alternateFormatData? same below for shortNumberMetadata; alternatively, make sure you make the same changes in our copy.

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
```diff
@@ -157,9 +130,8 @@ private static void loadShortNumberMetadataFromFile(String regionCode) {
       // that they are present, by checking the map of available data first.
       throw new IllegalStateException("missing metadata: " + fileName);
     }
-    PhoneMetadataCollection shortNumberData =
-        loadMetadataAndCloseInput(source, DEFAULT_BUFFER_SIZE);
-    for (PhoneMetadata metadata : shortNumberData.metadata) {
+    PhoneMetadataCollection shortNumberMetadata = loadMetadataAndCloseInput(source);
+    for (PhoneMetadata metadata : shortNumberMetadata.getMetadataList()) {
         regionCodeToShortNumberMetadataMap.put(regionCode, metadata);
```
lararennie: indentation looks 2 too far?

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
```diff
@@ -0,0 +1,1119 @@
+/*
+ * Copyright (C) 2010 The Libphonenumber Authors
```
lararennie: do we want to update the copyright? : D

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
```diff
@@ -0,0 +1,1119 @@
...
+      possibleNumberPattern_ = value;
+      return this;
+    }
+
+    // repeated int32 possible_length = 9;
+    private java.util.List<Integer> possibleLength_ = new java.util.ArrayList<Integer>();
+    public java.util.List<Integer> getPossibleLengthList() {
+      return possibleLength_;
+    }
+    public int getPossibleLengthCount() { return possibleLength_.size(); }
```
lararennie: so most methods in this file are Size, not Count: which do we want to do? which is the more "modern" way?

keghani: BuildMetadataFromXml.java refers to this method, so it needs to have the same name as the generated file for the C++ build. Are you asking about making the names in this file consistent long-term?

keghani: To answer your question - I don't know which is more "modern", no clue at which proto version this file was first created :)

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
```diff
@@ -0,0 +1,1119 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Definition of the class representing metadata for international telephone numbers. This class is
+ * hand created based on the class file compiled from phonemetadata.proto. Please refer to that file
```
keghani: This was written like this earlier (in 7.1.1). Do we really want to change it in this effort? I'm wary of making side fixes generally and more so in such large diffs.

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
keghani: No strong opinions from me, do you want us to :)? I would guess not, since I'm not the author and could not e.g. answer someone asking how this file was generated, e.g. at which version to run protoc to know what new methods should look like.

java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataManager.java
keghani: cl/131812609 :)

If you prefer I can change this here later. It's just more consistent with the file name constant and other names in this fie.

keghani: (I didn't want to sneak in side changes here, this was the way it was before but for some reason was renamed in summer of 2015).

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
lararennie: Yes, if the generated file has Count then we should probably at some point switch them all over :)

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
lararennie: I guess we can say it's the same copyright year, since we are resurrecting an old file. Maybe say in the commit which file it is based off.

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
lararennie: Sure a separate commit. But probably worth changing.

andyst:
LGTM

 - Merge request description (beginning) could be clearer in what we're doing / not doing. E.g. Does not affect (significantly): cpp version, JavaScript version. Removes dependency on protobuf (protobuf-nano) for Java version.

 - Does your testing (with part 2 / generated data) include verifying that we can still build cpp, JS and tests for cpp & JS pass?

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
keghani: I decided to do it in this commit but - I actually don't know how it was compiled!
Worth figuring this out and having it be consistent with the protobuf-java version in the C++ build tools if possible. I'll track separately.

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
keghani: What do you mean? Which file *is* it based off?
I mentioned in the commit that it's "almost what it was in 7.1.1", do you mean phonemetada.proto? I didn't generate it from that, I just rolled back and added new API by hand (mentioned in commit).

lararennie: I mean the Phonemetadata.java file that was there before we deleted it for nano.

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
keghani: Can do as part of the larger cleanup of C++ built tools and figuring out how to compile everything.
b/31538828

keghani:
Thanks! Addressed all comments (either took action in the new commit or left an open question or left an ack that will follow up in a later effort).

@andyst Yes :)

keghani:
@lararennie If/when you're happy please don't merge, I want to generate all the things (locally, not in the PR) and ensure tests pass with the latest changes.

java/libphonenumber/src/com/google/i18n/phonenumbers/Phonemetadata.java
keghani: I see :) Updated commit description.

keghani:
FTR - @lararennie mentioned happiness with PR so I'll merge momentarily.

PR #1342, opened 2016-09-15
Replace Tribe and Nextel with Virgin Mobile and WOM, respectively

kddlb:


kddlb:
I signed it!

kddlb:
Oh, I shall explain my changes:

- I replaced Tribe with Virgin Mobile, because the former is the company name, which is never seen by the carrier user (unless they read the contracts, or something.)
- And I replaced Nextel with WOM, because Nextel stopped existing with that name in July 2015.

PR #1351, opened 2016-09-19
Small test fixes (C++)

lararennie:
-- changing to use ULL in test numbers so this compiles on all platforms
-- updating a couple of test-cases/comments to bring inline with Java


keghani:
LGTM! I assume you ran tests :)
Please merge if you agree that we can push to master again.

lararennie:
Yes built and ran :)

PR #1355, opened 2016-09-20
Java possible numbers

lararennie:
Changing PhoneNumberUtil to use the possibleLengths information, not the reg-ex.


keghani:
LGTM pending discussion offline.

PR #1359, opened 2016-09-21
Porting usage of possibleLengths from Java to Javascript

lararennie:
Note that this shows the bug of "-1" appearing in possibleLengths for generalDesc, which it shouldn't: however, this only happens for a couple of countries that don't have fixed-line or mobile patterns (and hence NA or -1 is there) - and this is only a problem for the "withReason" variant, in that they are never tooShort, and it's a problem that must be live in Java already, so I think it's reasonable to fix in a separate commit.


keghani:
For me :) ?

keghani:
Thanks for the explanation! Agreed to fix that bug separately (even in another release is fine).
Reviewing commits...

javascript/i18n/phonenumbers/phonenumberutil.js
```diff
@@ -2770,11 +2771,18 @@ i18n.phonenumbers.PhoneNumberUtil.prototype.
  */
 i18n.phonenumbers.PhoneNumberUtil.prototype.isNumberMatchingDesc_ =
     function(nationalNumber, numberDesc) {
-
+  // Check if any possible number lengths are present; if so, we use them to
+  // avoid checking the validation pattern if they don't match. If they are
+  // absent, this means they match the general description, which we have
+  // already checked before a specific number type.
+
```
keghani: Optional: remove blank line

lararennie: Done

lararennie:
Jar changed because I rebuilt it to generate the js metadata, and it contained a couple of diffs from the Java possible lengths change.

PR #1364, opened 2016-09-23
Don't set empty preferred_domestic_carrier_code, and if we are passed such a number from clients then treat the empty field as if unset.

keghani:
** Begin note to reviewers**
When to merge? I'll wait until completion of release 7.1.1, then rebase and generate build tool jars with this code, and run tests.

Otherwise this is code-complete and ready for review.
** End note to reviewers**


lararennie:
LGTM.

PR #1372, opened 2016-09-28
Match short numbers in text

marado:
FindNumbers isn't currently able to find short numbers, which is a reasonable
expectation from it.

This patch changes the way phone numbers are verified, checking not only if
they are possible/valid numbers but also if they're possible/valid short
numbers.

Closes: #442

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -466,7 +466,8 @@ public class PhoneNumberUtil {
     POSSIBLE {
       @Override
       boolean verify(PhoneNumber number, String candidate, PhoneNumberUtil util) {
-        return util.isPossibleNumber(number);
+        ShortNumberInfo shortInfo = ShortNumberInfo.getInstance();
+        return (util.isPossibleNumber(number) || shortInfo.isPossibleShortNumber(number));
       }
     },
     /**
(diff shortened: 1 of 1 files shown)


keghani:
Thanks for taking a look at #442 @marado!

This task is unfortunately much bigger than one file as lararennie@ mentioned at #442.

Moreover PhoneNumberUtil.java does not support short numbers. This is also mentioned in our FAQ at [Why does PhoneNumberUtil return false for valid short numbers?](https://github.com/googlei18n/libphonenumber/blob/master/FAQ.md#why-does-phonenumberutil-return-false-for-valid-short-numbers)

We can't accept a pull request for #442 as there is significant work and testing involved for such a change, and reviewing it would be almost as much work as doing it ourselves. We can't prioritize this right now as metadata updates are prioritized over code changes, except for bug fixes; and we have quite a significant backlog.

I'll close this PR as we're tracking the issue already. As lararennie@ mentioned in the issue, you're welcome to build your own solution on top of PhoneNumberMatcher.java until we have the cycles to pick this up.

marado:
Thanks for your reply @keghani .

Would a patch that adds a new findNumbers method for ShortNumberInfo (that only looks for short numbers, but doesn't change any behaviour) be acceptable?

keghani:
Unfortunately we can't accept pull requests for large coding tasks at this time. The testing and the domain knowledge not to break existing use cases is rather large and we don't have the bandwidth to look into it right now. And we approach adding new API methods with caution and consideration as such methods cannot lightly be removed, so we can't just add a new functionality either. Sorry I don't have better news :(

PR #1391, opened 2016-10-18
add an extra falsehood: that phone numbers are numbers

DrHyde:
Oh god the data corruption horror that I have had to deal with when people forgot this.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1391)


FALSEHOODS.md
```diff
@@ -235,3 +235,10 @@ surprising how many false assumptions programmers continue to make about them.
     during, or after the actual implementation of numbering plan changes in the
     real world. The actual date on which a phone number range becomes active may
     not always adhere to official announcements.
+
+1.  **Phone numbers are numbers**
```
lararennie: Possible alternative titles:
"Phone numbers can be stored as numeric data types"
"Phone numbers can be treated as normal numbers"

Positioning:
I would put this after the "Users will only store phone numbers" one, since it is related in that if a user isn't storing a phone number then you *really* shouldn't treat it like a number.

FALSEHOODS.md
```diff
@@ -235,3 +235,10 @@ surprising how many false assumptions programmers continue to make about them.
...
+
+    Never try to store phone numbers in an int or any other kind of numeric
```
lararennie: nit: in -> as?

(also, ints are too small :) but we probably don't want to get into that.)

FALSEHOODS.md
```diff
@@ -235,3 +235,10 @@ surprising how many false assumptions programmers continue to make about them.
...
+    data type. You can't do arithmetic on them, and while 007, 07 and 7 are the
+    same number they are not necessary the same phone number. In some countries
+    a leading 0 is significant and forms part of the number itself.
```
lararennie: The last sentence should link to the falsehood we already have - A leading zero in numbers formatted for domestic usage can always be discarded when dialing from abroad

Also you should say "Moreover, a phone number may contain other diallable characters (see Phone numbers contain only digits) or an extension portion, dialled after waiting for a tone.

Without these points, I think people could think that storing the number in E164 format would work, since no phone number country calling code starts with a 0, and the + can be assumed.

FALSEHOODS.md
DrHyde: I agree about the positioning, so I'll move it. I prefer my title though - it's short and snappy, like in the list that (I assume) inspired this file: https://www.kalzumeus.com/2010/06/17/falsehoods-programmers-believe-about-names/

FALSEHOODS.md
DrHyde: good catch!

FALSEHOODS.md
DrHyde: Agreed.

FALSEHOODS.md
```diff
@@ -226,6 +226,17 @@ surprising how many false assumptions programmers continue to make about them.
...
+1.  **Phone numbers are numbers**
+
+    Never try to store phone numbers as an int or any other kind of numeric
+    data type. You can't do arithmetic on them, and while 007, 07 and 7 are the
+    same number they are not necessarily the same phone number - in some
+    countries a leading 0 is significant and forms part of the number itself
+    (see A leading zero in numbers formatted for domestic usage can always be
+    discarded when dialing from abroad). Moreover, a phone number may contain
+    other diallable characters (see Phone numbers contain only digits) or an
+    extension portion, dialled after waiting for a tone.
```
lararennie: We could also have the same tip here as line 173.

FALSEHOODS.md
```diff
@@ -226,6 +226,17 @@ surprising how many false assumptions programmers continue to make about them.
...
     to be a phone number, it should be stored as-is as entered by the user.
 
+1.  **Phone numbers are numbers**
+
+    Never try to store phone numbers as an int or any other kind of numeric
+    data type. You can't do arithmetic on them, and while 007, 07 and 7 are the
+    same number they are not necessarily the same phone number - in some
+    countries a leading 0 is significant and forms part of the number itself
+    (see A leading zero in numbers formatted for domestic usage can always be
+    discarded when dialing from abroad). Moreover, a phone number may contain
```
lararennie: I wish we could link easily to other parts of the doc; not sure how to do this though, so instead we could put the bit to see in italics just to make it read a bit better.

FALSEHOODS.md
lararennie: I see your point too, so ... okay :)

FALSEHOODS.md
DrHyde: Makes sense. There's one other (see blahblahblah) in the file so I'll italicize that too.

FALSEHOODS.md
DrHyde: OK

FALSEHOODS.md
lararennie: The other one isn't really needed, since it's actually a hyperlink (the xkcd one); but I don't care so to end the back and forth I will just accept this :)

lararennie:
Thank you very much for your contribution and for your prompt responses to comments!!!

PR #1407, opened 2016-10-26
Using possible length info in ShortNumberInfo

lararennie:
Using new possible length information to calculate whether a short number is the right length or not. This made us realise that the data was missing for MY; this was added.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1407)


cpp/src/phonenumbers/shortnumberinfo.cc
```diff
@@ -81,16 +81,19 @@ const PhoneMetadata* ShortNumberInfo::GetMetadataForRegion(
...
+// worth keeping this performance optimization.
 bool MatchesPossibleNumberAndNationalNumber(
-    const MatcherApi& matcher_api, const string& number,
-    const PhoneNumberDesc& number_desc) {
-  return matcher_api.MatchesPossibleNumber(number, number_desc) &&
-         matcher_api.MatchesNationalNumber(number, number_desc, false);
+    const MatcherApi& matcher_api,
+    const string& number,
+    const PhoneNumberDesc& desc) {
+  if (desc.possible_length_size() > 0 &&
```
keghani: Flagging as possibly needing to change due to offline discussion

keghani: Also below...

java/pending_code_changes.txt
```diff
@@ -1 +1,5 @@
-
+Code changes:
+ - Using new possibleLengthInfo to decide whether a short number is the right
+length or not. This could result in more specific results; whereas before, a
```
keghani: Please indent wrapped line to start at the same lace as "Using".

keghani: What impact does this have on the versioning scheme? Could argue for x.0.0 or x.y.z - what did we do for regular numbers?

lararennie: I would do x.y.z because it's like a metadata change: no one can or should change code because of it.

keghani: Ack, thanks!

cpp/src/phonenumbers/shortnumberinfo.cc
lararennie: Changed all places (I think)

PR #1421, opened 2016-11-05
fix timezones for 380

burunduk:
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1421)


lararennie:
Thanks! However, we need an explanation for any change to make it clear what the problem is and what the fix is. This is probably better suited as an issue rather than a PR.
Moreover, this file is auto-generated so we don't usually accept fixes like this to them.
Could you file an issue explaining the issue instead?

burunduk:
Ok, created the issue #1425 .
Offtopic: Are Bucharest has so many codes?

keghani:
Thanks @burunduk! I'll cose this PR since we're tracking this in an issue.

PR #1411, opened 2016-10-31
Move common metadata loading logic into MetadataManager, removing synchronization everywhere

keghani:
** Note for reviewers, do not copy into final commit **
cl/137686080
b/30158468
b/30807096

pkgdiff:
~keghani/GitHub_diffs/f51e431/pkgdiff_reports/cpp-build/1.0-SNAPSHOT-jar-with-dependencies_to_1.0-SNAPSHOT-jar-with-dependencies/changes_report.html
~keghani/GitHub_diffs/f51e431/pkgdiff_reports/java-build/1.0-SNAPSHOT-jar-with-dependencies_to_1.0-SNAPSHOT-jar-with-dependencies/changes_report.html

The jars picked up some .class changes since https://github.com/googlei18n/libphonenumber/pull/1432.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1411)

PR #1433, opened 2016-11-14
Update metadata.cc due to removing references to possible number pattern field

keghani:
*** do not copy into final review ***

+cc @andyst 

Discovered when re-generating files after merging #1432 into #1411.

This makes sense, because `lite_metadata.cc` was updated in #1432 and `metadata.cc` should contain all that metadata plus example numbers.

Steps to check that `metadata.cc` wasn't also updated in  #1432:
*   The commit in master corresponding to #1432 is df44ac6
*   So in a local client, I ran `git diff df44ac6^ df44ac6 > out`
*   Then examined `out`; the only hits matching `metadata.cc` are:
    ```
    lite_metadata.cc
    short_metadata.cc
    test_metadata.cc
    ```

Steps to generate all cpp metadata and run tests, for future reference:
*   From root of repository with updated tools jars:
    ```
    mkdir -p cpp/build/ &&\
    ( cd cpp/build/ && cmake -DUSE_LITE_METADATA=OFF .. && make ) &&\
    ( cd cpp/build/ && cmake -DUSE_LITE_METADATA=ON .. && make ) &&\
    ( cd cpp/build/ && cmake .. && make && ./libphonenumber_test )
    ```

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1433)


lararennie:
https://reviewable.io/reviews/googlei18n/libphonenumber/1432 also shows a better list of files.

Thanks!!

keghani:
> https://reviewable.io/reviews/googlei18n/libphonenumber/1432 also shows a better list of files.

Ah, yes! thanks :)

PR #1435, opened 2016-11-17
Loading cleanups

keghani:
** do not copy into final commit **

cl/139563478
cl/139572904

pkgdiff:
~keghani/GitHub_diffs/PR1435/pkgdiff_reports/java-build/1.0-SNAPSHOT-jar-with-dependencies_to_1.0-SNAPSHOT-jar-with-dependencies/changes_report.html
No diff in cpp-build jar.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1435)


andyst:
Please update the commit message ("** do not copy into final commit **")

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -968,48 +958,48 @@ static synchronized void setInstance(PhoneNumberUtil util) {
...
    *
-   * @param metadataLoader Customized metadata loader. This should not be null.
-   * @return a PhoneNumberUtil instance
+   * @param metadataSource  customized metadata source. This should not be null
+   * @return  a PhoneNumberUtil instance
    */
-  public static PhoneNumberUtil createInstance(MetadataLoader metadataLoader) {
-    if (metadataLoader == null) {
-      throw new IllegalArgumentException("metadataLoader could not be null.");
+  private static PhoneNumberUtil createInstance(MetadataSource metadataSource) {
```
andyst: Can you summarize again the rationale why
 a) we can make this public method private without this being considered a backwards-incompatible API change?
 b) clients of our library only need access to createInstance(MetadataLoader) but not createInstance(MetadataSource)?

keghani:
Re. "Please update commit message", what do you mean? That's not the commit message, but a description of the PR.

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
andyst: Never mind. I see you summarize this in the release notes. Thanks.

keghani: (a) [MetadataSource](https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/MetadataSource.java) and all 2 of its implementations ([1](https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/MultiFileMetadataSourceImpl.java), [2](https://github.com/googlei18n/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/SingleFileMetadataSourceImpl.java)) are non-public so even when we say `public PhoneNumberUtil.createInstance(MetadataSource)` public usage cannot pass anything to this method. FWIW I put this in the pending changes too.

(b) Clients can call getInstance(), which under the hood will (the first time) do the same as calling createInstance(new MultiFileMetadataSourceImpl(DEFAULT_METADATA_LOADER)) - (any time after that it will use the already created instance). The alternative would be to use their own loader, which they can do via createInstance(MetadataLoader), or to specify SingleFileMetadataSourceImpl, which they programmatically cannot do, and which we don't support anyway (no jars with single file right now).

keghani: To clarify that last point, the second "alternative" is actually not possible right now, even with this method signature staying public. I'm not saying it wouldn't be good to have it public, or that it's not useful for clients, but it's not usable so the signature shouldn't say public.

keghani:
Grr, race conditions. Got it!

keghani:
Thanks! I'll merge after re-generating jars and merging changes from master etc.

PR #1438, opened 2016-11-18
Make tools consistent

keghani:
cl/139578080

pkgdiff says UNCHANGED for cpp-build and java-build.
There is a diff in MetadataFilter.class files in both jars, but that's it.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1438)


keghani:
Unsure why Travis is unhappy - I can't reproduce the failure locally.
Looks like the install plugin 2.4 is online:
http://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-install-plugin/2.4/
https://mvnrepository.com/artifact/org.apache.maven.plugins/maven-install-plugin/2.4

It is possible that the URL changed, or that something in Travis changed. I didn't change any build files but the error looks like it's related to the plugin, so I'll let this rest over the weekend then take another look.

keghani:
A test with a no-op change causing Travis to fail confirmed that this is not necessarily a failure in the PR: https://github.com/googlei18n/libphonenumber/pull/1439

andyst:
@Travis: Can you trigger it again? (I don't know how, but there's probably a way.)

keghani:
I couldn't figure out how to trigger a new CI run without actually changing the PR content.
I ended up squashing the 2 commits. The issue seems to be fixed, this run passed.

PTAL :)

PR #1459, opened 2016-11-29
Updating the JS code to throw an error rather than a "string"

lararennie:
Changes required by clients: If anyone is doing custom behaviour depending on the value of the error, they will have to now compare to the value of e.message.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1459)


lararennie:
Based on original PR https://github.com/googlei18n/libphonenumber/pull/587

andyst:
@throwing Error instead of string: Isn't that a backwards-incompatible API change?

Do we want to group it with other backwards-incompatible API changes?

lararennie:
We are already making the next release v8.0.0 due to other backwards-incompatible changes for Java and C++.

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataJsonFromXml.java
```diff
@@ -279,6 +279,32 @@ private static void toJsArray(PhoneNumberDesc desc, JSArrayBuilder jsArrayBuilde
...
+    if (possibleLengthSize > 0) {
+      jsArrayBuilder.beginArray();
+      for (int i = 0; i < possibleLengthSize; i++) {
+        jsArrayBuilder.append(desc.possibleLength[i]);
+      }
+      jsArrayBuilder.endArray();
+    } else {
+      jsArrayBuilder.append(null);
+    }
+    // repeated int32 possible_length = 10;
```
andyst: s/possible_length/possible_length_local/ ?

ruimarinho:
Thank you for finally adding this. Refs https://github.com/googlei18n/libphonenumber/pull/690.

PR #1526, opened 2017-01-26
Make normalizeDiallableCharsOnly public and port it to javascript,

lararennie:
making use of it in the code in parallel places to the Java
implementation.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1526)

### DIFF
--- a/java/pending_code_changes.txt
+++ b/java/pending_code_changes.txt
@@ -1,3 +1,4 @@
 Code changes:
- - Made normalizeDiallableCharsOnly() API in PhoneNumberUtil public. This
-   method is already public in the C++ implementation.
+ - Made normalizeDiallableCharsOnly() API in PhoneNumberUtil (Java) public. This
+   method is already public in the C++ implementation. It has also now been
+   added to the Javascript implementation.
--- a/java/pending_code_changes.txt
+++ b/java/pending_code_changes.txt
@@ -1,1 +1,3 @@
-
+Code changes:
+ - Made normalizeDiallableCharsOnly() API in PhoneNumberUtil public. This
+   method is already public in the C++ implementation.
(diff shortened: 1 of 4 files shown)


javascript/i18n/phonenumbers/phonenumberutil.js
```diff
@@ -1096,6 +1096,22 @@ i18n.phonenumbers.PhoneNumberUtil.normalizeDigitsOnly = function(number) {
...
+ * all non-ASCII digits).
+ *
+ * @param {string} number a string of characters representing a phone number.
+ * @return {string} the normalized string version of the phone number.
+ */
+i18n.phonenumbers.PhoneNumberUtil.normalizeDiallableCharsOnly =
+    function(number) {
+
+  return i18n.phonenumbers.PhoneNumberUtil.normalizeHelper_(number,
+      i18n.phonenumbers.PhoneNumberUtil.DIALLABLE_CHAR_MAPPINGS_, true);
```
andyst: add /* remove non matches */ for the boolean parameter? We have that in Java and had it in one of the call sites in JS too.

java/pending_code_changes.txt
```diff
@@ -1 +1,3 @@
-
+Code changes:
+ - Made normalizeDiallableCharsOnly() API in PhoneNumberUtil public. This
```
andyst: s/API in PhoneNumberUtil public./in PhoneNumberUtil public in Java API./

And I'd mention that the method was added to JS too.

javascript/i18n/phonenumbers/phonenumberutil.js
lararennie: Sure. We are a bit inconsistent since we don't do it in the next method in Java either (convertAlphaCharactersInNumber) but oh well.

java/pending_code_changes.txt
lararennie: I really feel we should move this file up a level since it looks like it only applies to Java changes and that's how I use it, but it would be more useful if we also mentioned JS etc....

PR #1567, opened 2017-02-10
Fix for ExactlySameAs (c++) and IsNumberMatch

lararennie:
-- ExactlySameAs in C++ didn't look at the number_of_leading_zeros field
-- isNumberMatch now ignores the number of leading zeros if italian_leading_zero is set to false, and considers explicitly set values that match the defaults to be the same as the defaults. Note this shouldn't affect any user who creates phone numbers using "parse" as per recommendations.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1567)


javascript/i18n/phonenumbers/phonenumberutil.js
```diff
@@ -966,19 +966,19 @@ i18n.phonenumbers.PhoneNumberUtil.ValidationResult = {
   /** The number length matches that of valid numbers for this region. */
   IS_POSSIBLE: 0,
   /**
-   * The number length matches that of local numbers for this region only (i.e. numbers that may
-   * be able to be dialled within an area, but do not have all the information to be dialled from
-   * anywhere inside or outside the country).
+   * The number length matches that of local numbers for this region only (i.e.
+   * numbers that may be able to be dialled within an area, but do not have all
+   * the information to be dialled from anywhere inside or outside the country).
```
lararennie: Fixed line length to make lint happy while here.

keghani:
There is an actual fix in here though (even for people correctly only creating PhoneNumbers via "parse").

ExactlySameAs - shouldn't that be in the release notes?

lararennie:
It's c++ only so not sure; release notes is in the Java folder so don't
think anyone would look there for C++ changes. Happy to add though if you
think in the "wrong" place is better than not at all!!

I think we should move the release notes and pending code changes files up
one level after the next release - what do you think? I think at some point
we didn't because it was only Java that has official release numbers since
it goes to maven but now the js demo refers to the same version number etc..

10. feb. 2017 4.58 p.m. skrev "Keghani Kouzoujian" <notifications@github.com
>:

There is an actual fix in here though (even for people correctly only
creating PhoneNumbers via "parse").

ExactlySameAs - shouldn't that be in the release notes?

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/googlei18n/libphonenumber/pull/1567#issuecomment-278982606>,
or mute the thread

keghani:
We've been adding all sorts of changes there so I'd mention it. Yes, let's move one level up!
They do end up in the git tag of that name so the C++ code does get versioned.

PR #1594, opened 2017-02-24
Add std=c++11 to CMakeLists for older compilers

Armagetron:
Compiling the library with older compilers like gcc 4.9 that do not compile the code with the C++11 standard by default may cause errors. Therefore I created this pull request that adds the C++11 standard explicitly.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1594)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -333,6 +333,8 @@ if (${USE_ICU_REGEXP} STREQUAL "ON")
   endif ()
 endif ()
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++11")
+
 if (NOT WIN32)
   add_definitions ("-Wall -Werror")
 endif ()


Armagetron:
I signed it!

lararennie:
Can you explain what problem this is fixing? We aren't expecting people to use c++11 compilers: are there parts of the code where c++11 has snuck in? You say "may cause errors" -> has it caused errors?

Armagetron:
I attached a [build log](https://github.com/googlei18n/libphonenumber/files/799513/log.txt) I get when I compile the current version (master HEAD) on a Raspberry Pi with Raspbian Jessie and a gcc 4.9.2. With the fix I provided the error is gone.



I think this is caused by the protobuf compiler which generates C++11 code by default.

lararennie:
Thanks for more context!
Indeed, the latest version seems to require C++11 by default (https://github.com/google/protobuf/issues/597)
However (this patch)[https://github.com/google/protobuf/commit/ffde972ac705e70f597267537b033d7c00ea54c5] seems to turn them off.

We should see if the latest protobuf version still has this problem: it seems they are considering branching and having a c++11 version and a non-c++11 version so they didn't intend to turn it on...

Armagetron:
I tested the current libphonenumber (1f69a99ce31df7b7f6c869bfd51dc5d63e96ccc5) together with the current protobuf (google/protobuf@f0a5c1033a7a46b1405a1ce8526d4d9aa6d38375) version with the gcc 4.9 I don't get any errors. So it works without c++11 again.

However if the protobuf team decides to move their future development to a c++11 based version, libphonenumber should probably go the same way so that the newest version can be used. A pre c++11 fork would only receive bug-fix support and on systems that ship the new version by default one would have to manage two versions.

lararennie:
Great to hear it works again!
It is a good point that the systems shipping the new version would have to manage two versions, but not everyone is ready to move to C++11 either, so we'd have to make such a move carefully and with input from our users.
I'll close this in the meantime as it's not needed right now. Thanks for following up!

Issue #1010, opened 2016-03-15
Non-ASCII (beyond 0x7F) bytes in the source lead to C4819 warning when compiled with Visual Studio on WIndows + CJK locale


jungshik:
Four source files in cpp version of the library have non-ASCII bytes (they're in UTF-8 representation of hyphens and other characters):

./phonenumbermatcher.cc : "(\[（［"    and others 
./phonenumberutil.cc  :   ＋   -x‐-―−ー－-
./asyoutypeformatter.cc  : Two quotation marks
./phonenumberutil.h   ( "Google Zürich")

MSVC does not offer a way to treat all the input source file as encoded in UTF-8. They're always interpreted as in the "code page" for non-Unicode apps. 

When compiled on Windows by Visual C++ with the default code page for non-Unicode applications set to one of CJK code pages (932, 936, 949, 950), non-ASCII characters in UTF-8  are not valid and MSVC emits C4819 warnings. 

This warning can be ignored in most cases (especially because they're all in comments), but on rare occasions this can lead to a real compile error.  Moreover, some projects strive to be as warning-free as possible (usually, 'treat warning as error' is turned ON). 

An upcoming version of Visual Studio (not yet released) will offer an option (long overdue) to treat all the input files as encoded in UTF-8 (without 'UTF-8 BOM' at the beginning). 

It might be considered to either document this or to replace those non-ASCII characters in comments with U+hhhh notation. Perhaps, just documenting should be good enough. 

This is spun off from  https://bugs.chromium.org/p/chromium/issues/detail?id=454858

padmaksha:
@jungshik 

Thank you for taking the time to improve libphonenumber!

Could you please explain what exactly the documentation should contain and where it should be placed (e.g. in the CMakeLists file? README?)

keghani:
Hi @jungshik, open-source community,

I'm documenting the caveat in #1597.

To actually require the VS 2015 update 2 or later seems infeasible given our
bandwidth and expertise, but if you the reader have these then please
feel free to re-open this issue and take it up.

We don't actively maintain windows builds but would be able to accept a PR.
We'd also need someone else in the open-source community to test it
out and help with the review. Please see the [known Windows
issues](FAQ.md#what-about-windows), maybe someone else there or in the
[discussion group](https://groups.google.com/group/libphonenumber-discuss) could
help.

PR #1597, opened 2017-02-28
Update Visual Studio documentation

keghani:
Addresses #1010 
b/27688547

This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1597)

### DIFF
--- a/cpp/README
+++ b/cpp/README
@@ -263,6 +263,20 @@ Troubleshooting CMake via ccmake
 
 Building the library on Windows (Visual Studio)
 -----------------------------------------------
+We recommend that libphonenumber on Windows be built with Visual Studio 2015
+Update 2 (https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update2-vs)
+or later.
+
+This enables support of plain UTF-8 source files, for which you also need to specify
+the /utf-8 option in build files; see "New Options in VS2015 Update 2" at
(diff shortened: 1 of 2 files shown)


cpp/README
```diff
@@ -263,6 +263,20 @@ Troubleshooting CMake via ccmake
...
+Update 2 (https://www.visualstudio.com/en-us/news/releasenotes/vs2015-update2-vs)
+or later.
+
+This enables support of plain UTF-8 source files, for which you also need to specify
+the /utf-8 option in build files; see "New Options in VS2015 Update 2" at
+https://blogs.msdn.microsoft.com/vcblog/2016/02/22/new-options-for-managing-character-sets-in-the-microsoft-cc-compiler/,
+or a Chrome patch for an example:
+https://codereview.chromium.org/2488853002/diff/80001/build/config/compiler/BUILD.gn
+
+Without this version and the option, your compilation will likely break when
```
lararennie: We only have them in comments, because we made it explicitly not break, so I would say "your compilation will have warnings when ...."
From what I remember, we also made the comments the // kind -> /* style risks breakage because you need the */ to be correctly detected.

Unless there are files elsewhere where it matters, e.g. in the geocoding source code which was contributed so I don't know so well, or we changed policy at some point?

keghani: I don't know about a policy. I was paraphrasing best as I could from jungshik's explanations, I still don't have VS so don't appreciate the nuances :) I updated the language, WDYT?

PR #1623, opened 2017-03-10
Improve visibility for method: hasValidCountryCallingCode

jiafu1115:
we have telephone number's usage by format with country code-national numbers: The country code can be some numbers called internal country code we customized which not belong to standard ISO country codes. So we need one method to judge if the country code belong to ISO country codes and then to do some special logic.

So I found this method with private. so I advice to change to public for this method. Maybe it will be popular method to judge is country code is valid.

WDTY?

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1623)

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -1060,7 +1060,7 @@ public class PhoneNumberUtil {
   /**
    * Helper function to check the country calling code is valid.
    */
-  private boolean hasValidCountryCallingCode(int countryCallingCode) {
+  public boolean hasValidCountryCallingCode(int countryCallingCode) {
     return countryCallingCodeToRegionCodeMap.containsKey(countryCallingCode);
   }
 


lararennie:
I am concerned about what you are doing here.
- to format, you need a PhoneNumber object
- to get one, you must parse some text calling "parse"
- if you do this with an invalid country calling code, we will throw an exception.

Sounds like you aren't calling parse, in which case you are highly likely to be corrupting data, e.g. for Italian fixed-line numbers.

The only case where I could see this method being useful is for formerly valid country calling codes that no longer are, but in that case I would call isPossibleNumberWithReason on the number instead anyway since it'll tell you if the number seems plausible and if not why not. If the number is way too long or short, formatting it won't give you a very sensible output either....

In any case, you can just call  getRegionCodeForCountryCode and if it is ZZ (unknown), you know it's not a valid one.

Other alternatives: you can also getSupportedRegions and get the country calling code for each one and build the list yourself. Then you can also decide whether you want to include non-geographical entities too like international freephone, with getSupportedGlobalNetworkCallingCodes() which will return those.

jiafu1115:
@lararennie  Thanks for your feedback.
As you said, I doesn't want to parse phone number but just to judge if the country code is standard ones.
I can use your mentioned methods to write one method to meet requirements such as: 
```
boolean hasValidCountryCallingCode(int countryCallingCode){
    String regionCode = getRegionCodeForCountryCode(int countryCallingCode)
    return !("ZZ".equals(regionCode )
}
```
but if change to public for hasValidCountryCallingCode() in PhoneNumberUtil.  it will avoid everyone need this requirement write the duplicated similar method.
WDYT?

lararennie:
If you do not call parse, you cannot call format and expect that there will be no data corruption.

No one has requested this before, so I can't see evidence for increasing our API surface exposure when it is a one-liner in client code, and I can't see a valid use-case for doing this anyway - we already have more use-case based methods.

jiafu1115:
@lararennie total agree with you. Thanks.

PR #1627, opened 2017-03-14
miss 991: Trial of a proposed new international service, shared code

jiafu1115:
miss 991: Trial of a proposed new international service, shared code, 
refer to https://www.itu.int/dms_pub/itu-t/opb/sp/T-SP-E.164C-2011-PDF-E.pdf

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1627)


lararennie:
Again: we never change this file manually, it's auto-generated. We're unlikely to support this country calling code as we don't know of any numbers that actually connect/are valid. Please see the FAQ, specifically the "metadata" section and questions about the countries we support (https://github.com/googlei18n/libphonenumber/blob/master/FAQ.md#why-are-bouvet-island-bv-pitcairn-island-pn-antarctica-aq-etc-not-supported)

PR #1628, opened 2017-03-14
miss 388: Group of countries, shared code

jiafu1115:
miss 388: Group of countries, shared code, refer to https://www.itu.int/dms_pub/itu-t/opb/sp/T-SP-E.164C-2011-PDF-E.pdf

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1628)


lararennie:
Thanks! However, this file is auto-generated. If you think we are missing a country with this country calling code please file an issue instead and if you're right we'll update the xml file.

jiafu1115:
@lararennie Thanks.

Issue #1630, opened 2017-03-14
miss 388: Group of countries, shared code


jiafu1115:
refer to https://www.itu.int/dms_pub/itu-t/opb/sp/T-SP-E.164C-2011-PDF-E.pdf

 @lararennie

penmetsaa:
@jiafu1115 ,
We do not support all the country codes mentioned in ITU documents unless there are numbers assigned under that country code / do have numbering plan. Please take a look at our 
[FAQ page #why-are-bouvet-island-bv-pitcairn-island-pn-antarctica-aq-etc-not-supported](https://github.com/googlei18n/libphonenumber/blob/master/FAQ.vmd#why-are-bouvet-island-bv-pitcairn-island-pn-antarctica-aq-etc-not-supported). Please provide us supporting evidences. See our [guidelines](https://github.com/googlei18n/libphonenumber/blob/master/CONTRIBUTING.md#provide-supporting-evidence).

jiafu1115:
@penmetsaa  so miss 388 is due to can't make one valid metadata description for the 388.
BTW, one question for it? where is the source file to generate /com/google/i18n/phonenumbers/data/PhoneNumberMetadataProto_800. I can't find it.
And is there possible to support 388 without one metadata proto. Then at least we can know it is valid country code(though it is shared by countrys).
Thanks

penmetsaa:
(1) The source data is the main xml file  ([PhoneNumberMetadata.xml](https://github.com/googlei18n/libphonenumber/blob/master/resources/PhoneNumberMetadata.xml).

(2) Yes, we don’t have 388 in our phone metadata ([PhoneNumberMetadata.xml](https://github.com/googlei18n/libphonenumber/blob/master/resources/PhoneNumberMetadata.xml)) as there is no phone numbering plan available for it.  We auto-generate CoutryCodeToRegionCodeMap.java based on our phone metadata (PhoneNumberMetadata.xml) as we list only the region codes that we do have support for.

You can do a code change in your own port similar to your pull request #1628 but this is not recommended as you need to maintain your port up to date (in order to be up to date with phone numbering changes) and take care of the build as well - when we add new countries you have to merge.

What problem is this causing for you? Do you have diallable numbers with this country calling code?

jiafu1115:
[...]

PR #1635, opened 2017-03-16
avoid the similar issues #1630 or questions

jiafu1115:
avoid the similar issues #1630 or questions

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1635)


lararennie:
We don't list all countries or country codes we don't support here, hence "etc".

If you think we should list them explicitly I'd rather do it in the body of the text, not change the title, plus do all of them.

Do you want to make the update (and sync) or should we take over?

lararennie:
Replaced by PR #1681 since we didn't hear back.

PR #1637, opened 2017-03-16
improve performance by judging leading digits

jiafu1115:
improve performance by judging leading digits

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1637)

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -2192,6 +2192,12 @@ public class PhoneNumberUtil {
       return false;
     }
     String nationalSignificantNumber = getNationalSignificantNumber(number);
+    if (metadata.hasLeadingDigits()) {
+        if (regexCache.getPatternForRegex(metadata.getLeadingDigits())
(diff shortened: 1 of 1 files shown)


jiafu1115:
this is not only performance improve but fix one bug:

```
		PhoneNumber phoneNumber = instance.parse("+262269123456", null);
		String region = instance.getRegionCodeForNumber(phoneNumber); //YT
		System.out.println(region);
		System.out.println(instance.isValidNumberForRegion(phoneNumber, region));  //expect true,   actural: false
```

lararennie:
This is not a bug. This is not a valid number for this region.
The most we could do is a quick exit here (i.e. if the number doesn't start with the leading digits, exit with false). I'm not convinced it would be much of a performance win though.
I also note our leading digits for Mayotte don't even *include* all the valid prefixes (e.g. 80) - this is because the number is also valid elsewhere, and the leading digits are for things which are unique to that region. Note that VC for instance has "784", so if we used the code as an early exit we would invalidate all toll-free and premium-rate numbers for this region.

Closing this PR as it doesn't work the way it is, and it won't work even as an early exit.

lararennie:
Please see the comment in phonemetadata.proto, notably "This is not intended as a replacement for IsValidForRegion, and does not mean the number must come from this region (for example, 800
numbers are valid for all NANPA countries.)"

Issue #1642, opened 2017-03-17
Allow for alternate MetadataManager implementations


tadfisher:
On Android, `ClassLoader#getResourceAsStream` is slow and prone to concurrency issues. See this stacktrace snippet for an example of the latter:

```
Caused by java.util.ConcurrentModificationException
       at java.util.HashMap$HashIterator.nextEntry(HashMap.java:787)
       at java.util.HashMap$KeyIterator.next(HashMap.java:814)
       at java.util.jar.JarVerifier.readCertificates(JarVerifier.java:266)
       at java.util.jar.JarFile.getInputStream(JarFile.java:380)
       at libcore.net.url.JarURLConnectionImpl.getInputStream(JarURLConnectionImpl.java:222)
       at java.net.URL.openStream(URL.java:470)
       at java.lang.ClassLoader.getResourceAsStream(ClassLoader.java:444)
       at java.lang.Class.getResourceAsStream(Class.java:1169)
       at com.google.i18n.phonenumbers.MetadataManager$1.loadMetadata(MetadataManager.java:53)
       at com.google.i18n.phonenumbers.MetadataManager.getMetadataFromSingleFileName(MetadataManager.java:185)
       at com.google.i18n.phonenumbers.MetadataManager.getMetadataFromMultiFilePrefix(MetadataManager.java:116)
       at com.google.i18n.phonenumbers.MultiFileMetadataSourceImpl.getMetadataForRegion(MultiFileMetadataSourceImpl.java:64)
       at com.google.i18n.phonenumbers.PhoneNumberUtil.getMetadataForRegion(PhoneNumberUtil.java:2127)
       at com.google.i18n.phonenumbers.PhoneNumberUtil.parseHelper(PhoneNumberUtil.java:2986)
       at com.google.i18n.phonenumbers.PhoneNumberUtil.parseAndKeepRawInput(PhoneNumberUtil.java:2878)
       at com.google.i18n.phonenumbers.PhoneNumberUtil.parseAndKeepRawInput(PhoneNumberUtil.java:2867)
```

It would be very useful to allow for an alternate MetadataManager implementation which can load metadata from arbitrary locations, such as a file shipped as an Android asset in an APK.

Other libraries like Joda-Time and ThreeTenBP have similar issues, which has resulted in Android-specific libraries which wrap and replace their metadata-loading implementations:

- https://github.com/dlew/joda-time-android
- https://github.com/JakeWharton/ThreeTenABP

[...]

PR #1643, opened 2017-03-17
Add a no-metadata JAR variant

tadfisher:
Ref #1642 

This adds a JAR target without included metadata. Third parties using this
variant may supply their own metadata via a custom MetadataLoader.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1643)

### DIFF
--- a/java/libphonenumber/pom.xml
+++ b/java/libphonenumber/pom.xml
@@ -42,11 +42,32 @@
       <plugin>
         <artifactId>maven-jar-plugin</artifactId>
         <version>3.0.2</version>
-        <configuration>
-          <archive>
-            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-          </archive>
-        </configuration>
+        <executions>
+          <execution>
+            <id>default-jar</id>
+            <configuration>
+              <archive>
+                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+              </archive>
+            </configuration>
+          </execution>
+          <execution>
+            <id>no-metadata</id>
+            <phase>package</phase>
+            <goals>
+              <goal>jar</goal>
+            </goals>
+            <configuration>
+              <classifier>no-metadata</classifier>
+              <archive>
+                <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
+              </archive>
+              <excludes>
+                <exclude>com/google/i18n/phonenumbers/data/*</exclude>
+              </excludes>
+            </configuration>
+          </execution>
+        </executions>
       </plugin>
       <plugin>
         <groupId>org.codehaus.mojo</groupId>


keghani:
Thanks for the pull request @tadfisher!

Will this upload the no-metadata jar to Maven Central? Or will it be a local artifact only?
If the latter, why add this complexity instead of working on the regular jar? Some local work will need to happen anyway, to copy the data files over to the APK, and isn't it better to get those from the jar?

E.g. stripping data from the jar (quick hack, should probably use jar commands and probably doable wtihout the tempdir):

```shell
mkdir -p /tmp/tempdir
unzip -oq libphonenumber-x.y.z.jar -x com/google/i18n/phonenumbers/data/* -d /tmp/tempdir
pushd /tmp/tempdir > /dev/null
zip -oqr libphonenumber-no-metadata.x.y.z.jar .
rm -rf /tmp/tempdir
```

tadfisher:
In a third-party wrapper, it would be much better to depend directly on a Maven artifact rather than repackaging the jar.

- Consumers may freely replace the wrapper with the metadata-included dependency in unit tests or cross-platform build variants.
- Consumers may freely override the transitive dependency with a later version, not having to wait for the wrapper to catch up.
- Having a metadata-less base artifact reduces parallel work if someone wanted to change the metadata-loading facility for another use case, for example loading the rules from a database or the network.

Yes, that still means copying the metadata from the main artifact when building the wrapper, but that is a compile-only dependency for the wrapper, not a runtime dependency for the consumer.

tadfisher:
See also https://github.com/JodaOrg/joda-time/issues/220 and https://github.com/ThreeTen/threetenbp/issues/29 for similar use cases.

keghani:
Thanks for the details @tadfisher! I'll discuss with the team and get back to you this week.

java/libphonenumber/pom.xml
```diff
@@ -42,11 +42,32 @@
...
         <artifactId>maven-jar-plugin</artifactId>
         <version>3.0.2</version>
-        <configuration>
-          <archive>
-            <manifestFile>${project.build.outputDirectory}/META-INF/MANIFEST.MF</manifestFile>
-          </archive>
-        </configuration>
+        <executions>
+          <execution>
+            <id>default-jar</id>
```
keghani: Should this have a `phase` and `goals`, and a `classifier` in the `configuration`, to be consistent with the `execution` below?

tadfisher: I followed the [maven-jar-plugin example](https://maven.apache.org/plugins/maven-jar-plugin/examples/attached-jar.html); I don't believe these are necessary to add for the default jar configuration, but are required for adding an attached classified jar.

tadfisher:
@keghani Regarding the maven publishing question: I ran `mvn release:prepare -DdryRun=true`, and it built and had me sign the no-metadata jar as well as the standard jar. So I think you're good there. I don't see any artifact-specific configuration for the deploy or release plugin.

keghani:
Thanks! I patched this and ran some tests with and without the changes; the main libphonenumber jar didn't change as a result of this, and diffing that against no-metadata showed me data files that were only present in the main libphoneumber jar.

Thanks for verifying with the dry run! I actually didn't know that flag :)

Squashing!

Issue #1419, opened 2016-11-03
The demo site show carrier name as ??? when locale is fa


afarsi:
Hi there.
I checked the demo site of libphonenumber with fa locale. It can show English names of carriers. But It can't show carrier name when locale is fa(Persian/Farsi). You can check [here](https://libphonenumber.appspot.com/phonenumberparser?number=05112345678&country=IR&geocodingLocale=fa) and [here](https://libphonenumber.appspot.com/phonenumberparser?number=05112345678&country=IR&geocodingLocale=en) and compare Location Field of them.

I think it's because encoding of html page which is not UTF-8! Please fix it.

lararennie:
Yes, we know ;)  Thanks for the report; we'll try and fix this although our metadata updates have priority right now, so it might have to wait a short time.

keghani:
This was fixed with https://github.com/googlei18n/libphonenumber/issues/1646

PR #1646, opened 2017-03-19
Make demo show UTF-8 chars for a single result & fixing to work with mvn

lararennie:
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1646)

### DIFF
--- a/java/demo/pom.xml
+++ b/java/demo/pom.xml
@@ -1,66 +1,20 @@
 <?xml version="1.0"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
-  <parent>
-    <artifactId>libphonenumber-parent</artifactId>
-    <groupId>com.googlecode.libphonenumber</groupId>
-    <version>8.3.4-SNAPSHOT</version>
-  </parent>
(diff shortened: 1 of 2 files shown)


lararennie:
Fixes among other things issue #1419

lararennie:
Note: it is quite hard to run the demo at the moment, the README instructions using maven don't work, and using ant you have to manually copy the libs & remember to set the java version to 1.7 before you compile & copy the app-engine sdk into the right place.

Wondering if we want to update the README to at least explain the ant way.

lararennie:
Mm the maven problem (present with no change) seems to prevent the continuous build from working too :/

java/demo/pom.xml
```diff
@@ -1,67 +1,21 @@
 <?xml version="1.0"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
+  <groupId>com.googlecode.libphonenumber</groupId>
+  <artifactId>demo</artifactId>
+  <version>8.3.4-SNAPSHOT</version>
+  <packaging>jar</packaging>
```
andyst: So adding <packaging> and dropping the <repositories> did the trick?
Thanks for figuring that out!

lararennie:
The bit that fixed it seemed to be removing the "repositories" section. It seems to be bad practice to specify it unless you have your own private one (we don't).

Yes will update description!

SumathiSundara:
Fixed in release 8.4.0

PR #1650, opened 2017-03-20
isPossibleForType[WithReason] and getSupportedTypesFor*

lararennie:
New APIs:
-- isPossibleForType
-- isPossibleForTypeWithReason
-- getSupportedTypesForRegion
-- getSupportedTypesForNonGeoEntity

Added example numbers to the test metadata, updated a test accordingly.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1650)


lararennie:
-- Still TODO as part of this PR: porting the getPossibleForType methods to JS and adding tests for them, but since it's a big PR thought I'd send it for review now.
-- Note multiple commits so it's reviewable - many of the metadata files changed so makes the PR a little enormous.

Note that getSupportedTypesFor* methods need full metadata, not metadata-lite, maybe should document this somewhere - not sure where we document the restrictions if someone uses metadata-lite?

keghani:
Re. documenting metadata-lite restrictions, currently it's excluding example numbers and that's it. This PR doesn't change that, blacklist is the same. I don't have historical context into documenting that in GitHub.

lararennie:
No, but this CL introduces a new method which won't work without the
example numbers.

If we don't have any faq about this option we should add one. I'll do that.

20. mar. 2017 5.45 p.m. skrev "Keghani Kouzoujian" <notifications@github.com

java/libphonenumber/test/com/google/i18n/phonenumbers/CountryCodeToRegionCodeMapForTesting.java
```diff
@@ -31,10 +31,10 @@
   // countries sharing a calling code, such as the NANPA countries, the one
   // indicated with "isMainCountryForCode" in the metadata should be first.
   static Map<Integer, List<String>> getCountryCodeToRegionCodeMap() {
-    // The capacity is set to 33 as there are 25 different entries,
+    // The capacity is set to 36 as there are 27 different entries,
```
keghani: Weird that this increased by 3. I see only 2 new regions: TA and +882 (just wondering)

keghani: Oh never mind - I see the 25 became 27 so that's fine.

keghani: (It's a load factor thing)

keghani:
Jar diffs in this PR (from master) look as expected, but I still get some more diffs when I build locally. Maybe because some changes happened since they were created? (Maybe even in master that we brought back in?)

None of the metadata changes as a result of this though, so I'll merge. I might send something separately.

PR #1659, opened 2017-03-24
Update jars

keghani:
See diffs:

~keghani/GitHub_diffs/PR1659/

Please don't merge anything else in between (otherwise I should re-generate everything).

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1659)


keghani:
I suspect some of the .class diffs are because I compiled with Java 7.
But others are actual changes, e.g. descHasData (also seen in the .java changes).
The metadata file changes are because the metadata was re-generated after the build tools jar was re-generated (with it), then they end up in the jar when we build it again. It's not *used* in the jar but is in the same directory as a resource the jar uses (Phonemetadata.java), so just a side-effect of how we build right now.

PR #1671, opened 2017-03-31
Adding two more compilation warnings - extra and missing requires.

lararennie:
Added the missing requires this detected, removed the extra require. One
missing require was added manually (goog.object) since it's unable to
detect it.

Updated the readme: compiling everything seems to work at the moment.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1671)


javascript/README.md
```diff
@@ -69,9 +69,7 @@ project:
 
 4. Run the Closure Compiler to get your changes syntax and type checked. This will also generate `demo-compiled.js` used by `demo-compiler.html`
 
-  (TODO: Use target "compile" rather than "compile-demo" once visibility issues for getExtractedNationalPrefix_ have been resolved):
```
andyst: Is this TODO / note no longer needed? `ant compile` works now?

lararennie: Indeed, it works, even though the visibility issues haven't been resolved. If it stops working, I also know how to fix it - by moving the flags that care about visibility into the demo rule only, not the test rule.

PR #1679, opened 2017-04-07
Returning new validation types for ValidationResult

lararennie:
Returning IS_POSSIBLE_LOCAL_ONLY and INVALID_LENGTH from isPossibleNumberWithReason for local-only numbers and numbers that don't match any of the valid lengths for numbers in a particular country, but yet are not too short nor too long.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1679)


pending_code_changes.txt
```diff
@@ -1,3 +1,6 @@
 Code changes:
```
andyst: I'd reorder them to be most significant change first.

lararennie: Done

PR #1692, opened 2017-04-11
Update README.md

penmetsaa:
Update the release number that the demo is using

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1692)


lararennie:
Technically the description should be "update the release number that the demo is using" or something.

PR #1702, opened 2017-04-21
Fix duplicate variable initialization

keghani:
b/37535740

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1702)


keghani:
No worries! I thought about that too, I think it's because by the end of the initialization's try block it's definitely initialized since it couldn't fail on that line. So whether or not it gets to that catch, it's still in scope? Don't know enough about JS much less its compiler, but seems feasible.

PR #1706, opened 2017-04-25
Allow commas in extensions

voronaam:
Fixes issue #467 for Java.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1706)

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -304,8 +304,8 @@ public class PhoneNumberUtil {
   private static final String DEFAULT_EXTN_PREFIX = " ext. ";
 
   // Pattern to capture digits used in an extension. Places a maximum length of "7" for an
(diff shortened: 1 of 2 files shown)


voronaam:
I signed it! (the CLA)

voronaam:
Internal JIRA reference: ARCH-5875

keghani:
Thanks for the pull request @voronaam!

This is unfortunately not enough. As @lararennie mentioned at #467, we need to research if people are using commas to "wait", and we need to have a signal so that we don't allow commas when detecting numbers in text. We can't prioritize working on or reviewing this anytime soon, thanks for understanding.

voronaam:
I am "people" and work for a company whose end users use commas in extenstions for wait signal.
What kind of research would you need?

keghani:
E.g. whether some users (internationally) use this notation to indicate 2 separate numbers - see @lararennie's question at #467. We don't want to make a change in the library that would work for some users but break others.

voronaam:
I posted comments to issue #467 I think it is best to keep conversation in there.

PR #1708, opened 2017-04-27
Regenerate build tools jars with new changes

keghani:
These changes do not affect the way metadata is built. They're simply an artefact of the build tools' having to depend on the entire directories where some of their dependencies are located (e.g. Phonemetadata.java), so the jars change when anything in those directories changes.

I'm doing this now to clearly see what affects my upcoming changes will have.

This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1708)


keghani:
Diffs: ~keghani/GitHub_diffs/PR1708/

keghani:
(please ignore the JS build.xml file commits, they're due to b/37535483 but not affecting anyone else right now...)

andyst:
Looks like nothing changes (only a whitespace change in javascript/build.xml and I assume the .jar changes are no-op).

Do you want / need to merge this PR or what's the goal?

keghani:
Yes, I'd like it merged because upcoming changes will do actual changes to the build tools and I'd like the diff then to reflect no distracting info.

keghani:
The jar changes are at ~keghani/GitHub_diffs/PR1708/ (also commented above)

keghani:
more on goal in first comment - I guess maybe you didn't get an email because I edited it post-email :)

keghani:
Thanks!

PR #1712, opened 2017-05-02
No Area Code 310 for Canada

snowmen10:
Area Code 310 is for Los Angeles. Having 310\d{4} here will impose "+13101234" to be a valid Canadian phone number which shouldn't be the case.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1712)


snowmen10:
I signed it!

lararennie:

We are not accepting pull requests for such metadata updates at this time. See
our [contribution guidelines](https://github.com/googlei18n/libphonenumber/blob/master/CONTRIBUTING.md).

Instead, we'd like you to file an issue and explain why you don't think this is a valid Canadian number.

(For this case: see e.g. https://www.tripadvisor.ca/Travel-g153339-s605/Canada:Telephones.html and the section on Seven-digit phone numbers starting with 310)

PR #1719, opened 2017-05-03
Add FAQ about range overlaps

mizuni:
Specific examples about AR and MX provide since they are countries with varied lengths, formats and prefixes depending on context.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1719)


FAQ.md
```diff
@@ -201,6 +201,26 @@ We'd love to consume machine-readable numbering plan data (assigned ranges,
...
 to achieve this, please do so. Thanks!
 
+### Why is this number from Argentina (AR) or Mexico (MX) not identified as the right number type?
+
+Certain countries' mobile and/or fixed line ranges may overlap, which may make
+accurate identification impossible without additional and explicit context.
+For example, when calling a mobile line from a fixed line in Argentina,
+you need to dial 15 before the subscriber number, or 9 if you're calling
+from another country. Without these additional digits, you may end up calling
+a different endpoint belonging to a fixed line!
```
lararennie: This last part isn't true: because I don't think (at least anymore) that you have cases where ABCDEF is assigned to fixed-line and 9 ABCDEF is assigned to mobile. It's just that it doesn't work without the additional digits...

FAQ.md
```diff
@@ -201,6 +201,26 @@ We'd love to consume machine-readable numbering plan data (assigned ranges,
...
+
+Similarly, Mexico has different mobile prefixes needed when calling from a fixed
+line such as 044 when calling locally, 045 when calling from another state, and
+1 when dialing from another country.
+
+Moreoever, these countries have different possible lengths for area codes and subscriber
```
lararennie: Moreoever -> moreover

FAQ.md
```diff
@@ -201,6 +201,26 @@ We'd love to consume machine-readable numbering plan data (assigned ranges,
...
+numbers depending on the city, which further complicate matters (e.g. Buenos Aires is 11
+followed by eight digits, but Río Gallegos is 2966 followed by six digits).
+
+Despite all the aforementioned complexity, users may not provide their phone number
+with all the additional context unless explicitly asked.
```
lararennie: We need to go into more detail here I think. Add an extra sentence? We should say "For instance, since SMS messages can be sent in Argentina from a mobile phone without a prefix, the user may not supply the mobile prefix." And also cover the fact that "Our metadata is currently not fine-grained enough to detect when a user has omitted the mobile prefix, so we rely on a mobile prefix being present to correctly identify the type of phone number."

This part is the most important thing, and should go at the start.

FAQ.md
```diff
@@ -72,11 +72,11 @@ data errors.
 
 ### What types of phone numbers can SMSs be sent to?
 
-SMSs can be sent to `TYPE_MOBILE` or `TYPE_FIXED_LINE_OR_MOBILE` numbers. However,
+SMSs can be sent to `TYPE_MOBILE` or `FIXED_LINE_OR_MOBILE` numbers. However,
```
lararennie: TYPE_MOBILE -> MOBILE as well

Issue #1265, opened 2016-08-12
Asia/Kolkata vs Asia/Calcutta


coreygeesey:
The current response from libphonenumber when getting the timezone for an india number (country code: IN) is "Asia/Calcutta"

From my research, that city was renamed from Calcutta to Kolkata in 2001, and thus the timezone has been renamed to "Asia/Kolkata." This is also based on http://www.iana.org/time-zones

Is it possible to get it changed to Asia/Kolkata? Or are you guys pulling from a different data set?

Thank You

padmaksha:
@coreygeesey 

Thanks for taking the time to improve libphonenumber!
We've accepted this issue.

We auto-generate the timezones, so we'll see if we can make a generic fix to this case.
(will take some time- just to manage expectations).

Thanks

keghani:
We won't change the ID to `Asia/Kolkata` because we use [CLDR identifiers](http://www.unicode.org/cldr/charts/latest/supplemental/zone_tzid.html) which are stable. See their response at http://unicode.org/cldr/trac/ticket/9718.
The ID stays the same despite new aliases for this timezone, but you can use CLDR to find the localised name.

Thanks for reporting!

Issue #1133, opened 2016-05-27
Timezone Issue US area code 623


aaronjwhiteside:
http://libphonenumber.appspot.com/phonenumberparser?number=6235529111&country=US

Location:   Arizona
Time zone(s):   [America/Denver]

Timezone should be America/Phoenix.

Denver observes daylight savings time but Phoenix does not.

padmaksha:
@aaronjwhiteside

Thanks for taking the time to improve libphonenumber!
We've accepted this issue.

We auto-generate the timezones, so we'll see if we can make a generic fix to this case.
(will take some time- just to manage expectations).

phuongmn:
Any update on this issue?

padmaksha:
We're working on it. We'll update this thread when it's released.

kkeshava:
Fixed in release 8.3.1.

Issue #1131, opened 2016-05-27
Timezone issue (USA Areacode 551)


agilesolmw:
Country/region affected : US
Example number(s) affected : 551-220-1111
The phone number range(s) to which the issue applies : 551-XXX-XXXX
The type of the number(s) : ALL
The cost, if applicable : N/A
Supporting evidence

http://libphonenumber.appspot.com/phonenumberparser?number=5512201111&country=US
It print wrong output (Time zone(s) [America/Los_Angeles])
It's not a time zone for New Jersey.

padmaksha:
@agilesolmw

Thanks for taking the time to improve libphonenumber!
We've accepted this issue.

We auto-generate the timezones, so we'll see if we can make a generic fix to this case.
(will take some time- just to manage expectations).

kkeshava:
Fixed in release 8.3.1.

PR #1721, opened 2017-05-03
Add timezone mapper readme

keghani:
cl/154937783
b/29383121

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1721)

### DIFF
--- a/resources/timezones/README.md
+++ b/resources/timezones/README.md
@@ -1,10 +1,13 @@
 # Timezone Mapper
 
 The timezone mapper allows one to get a likely timezone for a particular phone
-number. For mobile phones which are associated with particular area codes, it
-returns the timezone of the area code; it does not track the user's current
-location in any way. This could be used to work out whether it is likely to be a
-good time to ring a user based on their provided number.
+number. The timezone returned is the canonical ID from [CLDR](
+http://www.unicode.org/cldr/charts/latest/supplemental/zone_tzid.html), not a
+localised name (or any other identifier). For mobile phones which are associated
+with particular area codes, it returns the timezone of the area code; it does
+not track the user's current location in any way. This could be used to work out
+whether it is likely to be a good time to ring a user based on their provided
+number.
 
 Code Location:
 [java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java](https://github.com/googlei18n/libphonenumber/blob/master/java/geocoder/src/com/google/i18n/phonenumbers/PhoneNumberToTimeZonesMapper.java)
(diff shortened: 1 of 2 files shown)


resources/timezones/README.md
```diff
@@ -0,0 +1,31 @@
...
+
+```
+PhoneNumberToTimeZonesMapper timeZonesMapper = PhoneNumberToTimeZonesMapper.getInstance();
+
+List<String> timezones = timeZonesMapper.getTimeZonesForNumber(phoneNumber);
+```
+
+## Contributing to the timezone metadata
+
+The timezone metadata is auto-generated with few exceptions, so we cannot except
```
lararennie: except -> accept

resources/timezones/README.md
```diff
@@ -0,0 +1,31 @@
+# Timezone Mapper
+
+The timezone mapper allows one to get a likely timezone for a particular phone
+number. For mobile phones which are associated with particular area codes, it
```
lararennie: We should mention that the timezone returned is the canonical ID from CLDR (and link to CLDR), not a localised name (or any other identifier) - and then below in the "contributing" part we can mention that IDs in CLDR don't change (or whatever Mark said that time we raised the bug).

keghani:
PTAL

resources/timezones/README.md
```diff
@@ -0,0 +1,40 @@
+# Timezone Mapper
+
+The timezone mapper allows one to get a likely timezone for a particular phone
+number. The timezone returned is the canonical ID from [CLDR](
```
hagbard: Optional: Perhaps:
... allows a likely timezone to be obtained for a given phone number.

resources/timezones/README.md
```diff
@@ -0,0 +1,40 @@
...
+PhoneNumberToTimeZonesMapper timeZonesMapper = PhoneNumberToTimeZonesMapper.getInstance();
+
+List<String> timezones = timeZonesMapper.getTimeZonesForNumber(phoneNumber);
+```
+
+## Contributing to the timezone metadata
+
+The timezone metadata is auto-generated with few exceptions, so we cannot accept
+pull requests. If we have an error please file an issue and we'll see if we can
+make a generic fix.
```
hagbard: I'm not sure we need ""and we'll see if we can make a generic fix."" as that should be true for all bugs.

resources/timezones/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+If making fixes in your own fork while you wait for this, build the metadata by
+running this command from the root of the repository (assuming you have `ant`
+installed):
+
+```
+ant -f java/build.xml build-timezones-data
+```
+
+Note that, due to our using CLDR timezone IDs which are stable, we do not change
```
hagbard: ""using CLDR timezone IDs which are stable"" maybe ""using stable CLDR timezone IDs""

resources/timezones/README.md
keghani: This is what we've said when explaining that timezone metadata is auto-generated: https://github.com/googlei18n/libphonenumber/issues/1265, https://github.com/googlei18n/libphonenumber/issues/1131, https://github.com/googlei18n/libphonenumber/issues/1133

It's different for timezone metadata (and maybe certain regions' geocoding) than any other bugs, because it's about making an automatic fix that fixes the issues and doesn't change anything else. So the "fix" is to the process, not the metadata; and it has to be generic, not about the given issue.

resources/timezones/README.md
keghani: Done

resources/timezones/README.md
keghani: Done

keghani:
@lararennie did you want to take another look?

keghani:
Done! Good idea, thanks

PR #1726, opened 2017-05-04
BuildMetadata updates

lararennie:
Changing the BuildMetadataFromXml to only set same_mobile_and_fixed_line_pattern and
national_prefix_optional_when_formatting if they differ from default values, and the domestic_carrier_code_formatting_rule, international_prefix and national_prefix_formatting_rule if there is
one.

Adding a test with some golden data for the BuildMetadataJsonFromXml.

Adding to the proto a missing explicit default false.

Updating a couple of methods in the Phonemetadata.java class to have
names that match those of the auto-generated Phonemetadata used by the
CPP build, now that we use those methods when building the metadata.

Fixed an issue with the ant build file (encoding) that stopped the open-source instructions for how to regenerate metadata from working.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1726)


andyst:
Did a careful review, just didn't find anything. :)

PR #1730, opened 2017-05-08
Js proto changes

lararennie:
- adding some @final annotations
- deleting field that isn't in the proto anymore
- adding an explicit default that's now in the proto file

Tested:

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1730)


lararennie:
Note this code that is now deleted was omitted in compiled code anyway since it wasn't referred to.

PR #1739, opened 2017-05-11
Make canBeInternationallyDialled API public in Java and CPP

penmetsaa:
PhoneNumberUtil.canBeInternationallyDialled() API is made public (in Java and CPP) now as data is available in our metadata now and it is useful for clients if they want to make sure they only allow internationally-accessible numbers for a particular use-case. In JS this API is already public.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1739)


penmetsaa:
Hi Lara,

We only added line "(which are currently all presumed to not be diallable from outside their country)." to our java doc.
There is no line "Note that, at the moment, this method does not handle short numbers." in JS doc as there is no short number  support. So I have not updated it.

lararennie:
Yep, but I would still add it, because it is still true :)

penmetsaa:
Ok. I see your point now. Updated the doc. PTAL

lararennie:
CPP build jar hasn't changed, which is expected, other than the manifest file saying you built it, not someone else :D but you can keep it here just for posterity if you want.
I'll approve, you can merge and sync.

penmetsaa:
SGTM to commit CPP jar as you mentioned.. Thanks for the info Lara :)

PR #1751, opened 2017-05-22
Stop setting the national number pattern to NA when no data exists

keghani:
Also make MatcherApi implementations consistent and increase their test coverage.

b/37494614
cl/154818919
cl/156729949
cl/156732397

Jar diffs at ~keghani/GitHub_diffs/PR1751/

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1751)


keghani:
FWIW I didn't generate in that order - I just had all the files in my client at the end, and picked them starting in alphabetical order to put in different commits for easier review.

cpp/test/phonenumbers/matcher_test.cc
```diff
@@ -0,0 +1,105 @@
...
+  PhoneNumberDesc CreateDesc(
+      const string& national_number_pattern) const {
+    PhoneNumberDesc desc;
+    if (!national_number_pattern.empty()) {
+      desc.set_national_number_pattern(national_number_pattern);
+    }
+    return desc;
+  }
+
+  void ExpectMatched(
```
lararennie: None of these methods need to be on the class? they can be just in the anonymous namespace above.

andyst:
Why combine the seemingly unrelated changes (matcher API and NA) in a single PR / code review?

cpp/test/phonenumbers/matcher_test.cc
keghani: Done

keghani:
Included matcher API here since the increased test coverage gives me better confidence about "NA". Sorry it got so big...

PR #1780, opened 2017-06-07
Update 92.txt

najamkhn:
Updated the list with the new codes here: https://en.wikipedia.org/wiki/Telephone_numbers_in_Pakistan#Mobile_telephone_numbers

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1780)


penmetsaa:
@najamkhn,
 
Thank you for taking the time to improve libphonenumber!
 
IR.21 documents by Zong and Telenor mobile operators for Pakistan, says that the complete 31X and 34X range belongs to each operator respectively. For this reason,  generic entries (i.e "9231|Zong" and "9234|Telenor") are added in carrier mapping. 
 
This file does not affect validation. If there are number ranges that you think are not in use that you would like removed for validation purposes, please file an issue about this. Note that the wikipedia article seems out of date and we can find online working numbers outside the prefixes listed there.
 
Closing this as working as intended from our end. If you would like to open a new issue, please share any official evidence you may have. See our [guidelines](https://github.com/googlei18n/libphonenumber/blob/master/CONTRIBUTING.md#provide-supporting-evidence).

PR #1805, opened 2017-06-22
Removing leading_zero_possible from metadata and code

lararennie:
It was confusing and unnecessary. Not every country that has short numbers beginning with a 0 had it, and it is not the only digit that could overlap with a national prefix and hence be interpreted incorrectly.

Keeping in proto file, but will remove there later.

This should not materially affect users of the library - the only place it was used was 
FormatInOriginalFormat, and only initially to try and avoid modifying the input number by removing/adding digits inadvertently. Now this is checked at the end of the method anyway.
However slight formatting differences with this method on invalid numbers starting with 0s may be noticed in some countries.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1805)


pending_code_changes.txt
```diff
@@ -1 +1,9 @@
-
+Code changes:
+ - Removing leading_zero_possible from the metadata and all the places it is
+   referenced in the build and prod code. Will be removed from the metadata
+   proto itself in a subsequente release. This should not affect users of the
```
keghani: subsequent (remove trailing e)

lararennie: Fixed.

lararennie:
Hopefully fixed jars.

PR #1821, opened 2017-06-30
[JS] Bug fix for getNationalSignificantNumber

lararennie:
Fixing bug where getNationalSignificantNumber for an empty phone number object would return the string null in Javascript.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1821)


javascript/i18n/phonenumbers/phonenumberutil.js
```diff
@@ -2430,10 +2430,13 @@ i18n.phonenumbers.PhoneNumberUtil.prototype.
...
     function(number) {
 
-  // If leading zero(s) have been set, we prefix this now. Note this is not a
-  // national prefix.
+  if (!number.hasNationalNumber()) {
+    return '';
+  }
   /** @type {string} */
   var nationalNumber = '' + number.getNationalNumber();
+  // If leading zero(s) have been set, we prefix this now. Note these are not a
```
keghani: these are --> this is?
alternatively, change the whole thing:

`we prefix these now. Note these do not make up a national prefix.`

this -> these
there are not a -> these do not make up a

lararennie: Changed the whole thing :)

lararennie:
Added pending_code_changes.txt too.

keghani:
Still LGTM. Thanks for remembering about the demo!

PR #1823, opened 2017-06-30
Update javadoc on isNumberGeographical

keghani:
cl/160633190

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1823)


keghani:
It's in JS but only the outer function, not the helper.
Added the notes and updated the documentation in JS, but I can't quickly enough add the other function (keeps failing and I've exhausted my debugging skills for today), and I'd argue nothing's made it more important now, these functions were already public.

lararennie:
If it's already there and public that's great, don't need the helper. Thanks for updating the docs though!!

PR #1841, opened 2017-07-07
Add isSmsServiceForRegion(PhoneNumber, RegionCode) API in C++ and Java libraries

keghani:
Also serialize is_sms_service in the short number metadata.
JS API is not affected because it doesn't have short numbers.

When auto-updating metadata, only the short number metadata in C++ was updated, because it uses protoc, which was always serializing (empty) is_sms_service in the regular, test, and alternate formats metadata.
Java metadata is serialized using a hand-crafted Phonemetadata.java file which was not serializing is_sms_service, so now all of the Java metadata (including that where the value is empty) is updated to serialize is_sms_service.

Jar diffs: ~keghani/GitHub_diffs/is_sms_service

cl/155168852
cl/161084153
b/28121231

This will cause a minor version number update.
Could also potentially have a porters-only heads-up in the release tag about the metadata format changing, similar to http://github.com/googlei18n/libphonenumber/releases/tag/v8.4.2

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1841)


cpp/src/phonenumbers/shortnumberinfo.h
```diff
@@ -162,6 +162,18 @@ class ShortNumberInfo {
       const PhoneNumber& number,
       const string& region_dialing_from) const;
 
+  // Given a valid short number, determines whether it is an SMS service
+  // (however, nothing is implied about its validity). An SMS service is where
+  // the primary or only intended usage is to receive and/or send text messages
+  // (SMSs). This includes MMS as MMS numbers downgrade to SMS if the other
+  // party isn't MMS-capable. If it is important that the number is valid, then
+  // its validity must first be checked using {@link #isValidShortNumber} or
+  // {@link #isValidShortNumberForRegion}. Returns false if the number doesn't
```
lararennie: Sorry if I missed this before, but we don't do {@link} in C++;
Method names should have capital 'I' too

pending_code_changes.txt
```diff
@@ -1,4 +1,10 @@
 Code changes:
+ - Added isSmsServiceForRegion(PhoneNumber, RegionCode) API in C++ and Java
+   libraries. An SMS service is where the primary or only intended usage is to
```
lararennie: I'd mention "short number info" libraries, or mention the class or file.

cpp/src/phonenumbers/shortnumberinfo.h
keghani: Done! Thanks for catching.
Yeah I missed it too because it was last-minute copying from Java. I had just made sure it made sense and voila!

pending_code_changes.txt
keghani: Done

PR #1846, opened 2017-07-12
Clarify distinctions between the two javascript ports

halt-hammerzeit:
Currently the README suggest two 3rd party javascript ports of this library.
It doesn't expose any differences though making it confusing and difficult for the end users to figure out which one should they choose.
I added some simple key points explaining the difference between the two.
@seegno tell us what you think.

---
This change is [Reviewable](https://reviewable.io/reviews/googlei18n/libphonenumber/1846)

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,8 +174,7 @@ maintenance processes.
 *   Javascript: If you don't want to use our version, which depends on Closure,
     there are several other options, including
     https://github.com/halt-hammerzeit/libphonenumber-js (a stripped-down
-    version) and https://github.com/seegno/google-libphonenumber (installable
-    via npm, a browserify-compatible wrapper)
+    version, about 110 KiloBytes in size) and https://github.com/seegno/google-libphonenumber (a browserify-compatible wrapper installable via npm, genuine all-in-one build of the original library, includes Google Closure library, about 420 KiloBytes in size)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -174,8 +174,7 @@ maintenance processes.
 *   Javascript: If you don't want to use our version, which depends on Closure,
     there are several other options, including
     https://github.com/halt-hammerzeit/libphonenumber-js (a stripped-down
-    version) and https://github.com/seegno/google-libphonenumber (installable
-    via npm, a browserify-compatible wrapper)
+    version, about 110 KiloBytes in size) and https://github.com/seegno/google-libphonenumber (a browserify-compatible wrapper installable via npm, genuine all-in-one build of the original library, includes Google Closure library, about 420 KiloBytes in size)
```
lararennie: Happy for more detail, but:
- KB not KiloBytes
- wrap at 80 chars
- what does "genuine" or "all-in-one" mean here? Does this add anything? I as a user am more confused now than before.

halt-hammerzeit:
@lararennie rewrote the differences description in other words.
I'm not a native English speaker so the choice of words may be not the best one but if you feel like you sense the direction you could suggest your own phrasing.

keghani:
Thanks @halt-hammerzeit !

PR #1849, opened 2017-07-13
Update GB (+44) carrier data from Ofcom

giggsey:
Uses the most recent carrier data from Ofcom, with carrier names being changed to friendlier versions (brand/trading names).

Data has also been compressed to reduce duplicate prefixes, although the script has compressed it further than it used to be.

Generated using https://github.com/giggsey/GBCarrierGenerator (so can be run again).

A few questions about the company names:
 - [x] The current data converts Hutchison 3G UK Ltd to Hutchison. Their commonly used brand name is 3/Three.
 - [x] Vectone Mobile Limited is called Mundio in your data (from Wikipedia: Vectone Mobile is operated by Mundio Mobile Ltd). However, their [website](https://www.vectonemobile.co.uk/) just calls themselves Vectone Mobile
 - [x] M P Tanner Limited t/a FIO Telecom has been translated to their trading name FIO Telecom

### DIFF
--- a/resources/carrier/en/44.txt
+++ b/resources/carrier/en/44.txt
@@ -13,9 +13,108 @@
 # limitations under the License.
 
 # Carrier details updated from http://www.ofcom.org.uk/static/numbering/S7.xls
-# Carrier names have been updated to friendlier trading/brand names
-# The mapping of company names can be found in the project below.
-# Generated on 2017-07-14 via https://github.com/giggsey/GBCarrierGenerator
+# Generated on 2017-07-14.
+
+# Carrier names have been updated to friendlier trading/brand names.
+# The mapping of company names can be found in the project below:
+#  - (aq) Limited trading as aql: aql
+#  - 08Direct Limited: 08Direct
+#  - 09 Mobile Ltd: 09 Mobile
+#  - 24 Seven Communications Ltd: 24 Seven
+#  - AMSUK LTD: AMSUK
+#  - Ace Call Limited: Ace Call
+#  - Airwave Solutions Ltd: Airwave
+#  - Alliance Technologies LLC: Alliance
+#  - Andrews & Arnold Ltd: Andrews & Arnold
+#  - Anywhere Sim Limited: Anywhere Sim
+#  - BT OnePhone Limited: BT OnePhone
+#  - Bellingham Telecommunications Limited: Bellingham
+#  - CFL Communications Limited: CFL
+#  - Cheers International Sales Limited: Cheers
+#  - Citrus Telecommunications Ltd: Citrus
+#  - Cloud9 Communications Limited: Cloud9
+#  - Cloud9 Mobile Communications Ltd: Cloud9
+#  - Compatel Ltd: Compatel
+#  - Confabulate Limited: Confabulate
+#  - Core Communication Services Ltd: Core Communication
+#  - Core Telecom Limited: Core Telecom
+#  - Dynamic Mobile Billing Limited: Oxygen8
+#  - EE Limited ( TM): EE
+#  - EE Limited (Orange): Orange
+#  - Esendex Limited: Esendex
+#  - FleXtel Limited: FleXtel
+#  - Fogg Mobile AB: Fogg
+#  - Gamma Telecom Holdings Ltd: Gamma Telecom
+#  - Global Reach Networks Limited: GlobalReach
+#  - Globecom International Limited.: Globecom
+#  - Guernsey Airtel Limited: Airtel
+#  - HAY SYSTEMS LIMITED: HSL
+#  - Hanhaa Limited: Hanhaa
+#  - Hutchison 3G UK Ltd: Three
+#  - IPV6 Limited: IPV6
+#  - IV Response Limited: IV Response
+#  - Icron Network: Icron
+#  - Icron Network Limited: Icron Network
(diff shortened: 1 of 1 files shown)


lararennie:
Thanks for this!
a) Please do change Hutchison to Three everywhere. or 3. Um. not sure which is best. Probably "Three". They say "About Three" on their website.
b) Fine to change Mundio to Vectone Mobile
c) Not sure what you think we should do for the last one?

In the file, can you add comments at the top for any translations you do from the data, and update the link to ofcom data if it has changed? A note about the compression technique used would be also helpful. We actually have an online app somewhere that can be used, but yours looks sane.

Also delete the empty blank line at the bottom thanks :)

giggsey:
Regarding item 3, it's more of a note instead of a question.

I'll make the changes, and update shortly (might be tomorrow).

As for the compression, do you have a link to that online tool? I'm slightly concerned that I compressed too hard. I sanity checked some ranges by hand, and they appear okay.

lararennie:
It looked okay to me :)
We have:
http://libphonenumber-combine-geodata.appspot.com/
I think it's even *more* aggressive than what you have though, there are definitely known issues about using it...

keghani:
It's too aggressive, we don't even use it for changes we make - and the last time I wished it was better was actually for this same file :)

E.g. it compresses
```
331|Paris
334|Paris
```
to
```
3|Paris
```

giggsey:
Updated Hutchison to Three and added a comment about how the data was generated.

I've also rebased to upstream's master.

resources/carrier/en/44.txt
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 # Carrier details updated from http://www.ofcom.org.uk/static/numbering/S7.xls
+# Carrier names have been updated to friendlier trading/brand names
+# The mapping of company names can be found in the project below.
```
lararennie: I would prefer an explicit mapping here, because your project may have a shorter lifetime than this one, who knows :) it's not so big?
Also if someone else contributes at any point, they may want to add a mapping.

resources/carrier/en/44.txt
```diff
@@ -13,6 +13,9 @@
 # limitations under the License.
 
 # Carrier details updated from http://www.ofcom.org.uk/static/numbering/S7.xls
+# Carrier names have been updated to friendlier trading/brand names
```
lararennie: and don't forget trailing full stop

resources/carrier/en/44.txt
giggsey: That's fine, but there are ~95 lines in my mapping array.

lararennie: 95?!
I guess we should still do it, they are only comments and will be stripped when built anyway, and it's good to verify them in one place.

giggsey: Updated with list embedded into the comments.

resources/carrier/en/44.txt
```diff
@@ -13,6 +13,108 @@
...
+#  - Guernsey Airtel Limited: Airtel
+#  - HAY SYSTEMS LIMITED: HSL
+#  - Hanhaa Limited: Hanhaa
+#  - Hutchison 3G UK Ltd: Three
+#  - IPV6 Limited: IPV6
+#  - IV Response Limited: IV Response
+#  - Icron Network: Icron
+#  - Icron Network Limited: Icron Network
+#  - JT (Guernsey) Limited: JT
+#  - JT (Jersey) Limited: Jersey Telecom
```
giggsey: I had just copied this over from the old data, but I think this should just be JT or Join Together (Source: http://www.jtglobal.com/Jersey/)

resources/carrier/en/44.txt
```diff
@@ -13,9 +13,108 @@
 # limitations under the License.
 
 # Carrier details updated from http://www.ofcom.org.uk/static/numbering/S7.xls
-# Carrier names have been updated to friendlier trading/brand names
-# The mapping of company names can be found in the project below.
-# Generated on 2017-07-14 via https://github.com/giggsey/GBCarrierGenerator
+# Generated on 2017-07-14.
+
+# Carrier names have been updated to friendlier trading/brand names.
+# The mapping of company names can be found in the project below:
```
lararennie: you don't mention the project now. Also I think it's fair to still reference your project in the way you did at line 18.

resources/carrier/en/44.txt
lararennie: Yep I agree I think JT is better.

resources/carrier/en/44.txt
```diff
@@ -13,9 +13,108 @@
...
+#  - Vectone Mobile Limited: Vectone Mobile
+#  - Virgin Mobile Telecoms Limited: Virgin Mobile
+#  - Vodafone Uk Ltd: Vodafone
+#  - Voicetec Systems Ltd: Voicetec
+#  - Voxbone SA: Voxbone
+#  - Wavecrest (UK) Ltd: Wavecrest
+#  - Yim Siam Telecom: Yim Siam
+#  - Ziron (UK) Ltd: Ziron
+#  - aql Wholesale Limited: aql
+
```
lararennie: remove extra blank line :)

giggsey:
Last comments addressed.

PR #1851, opened 2017-07-13
Add pull request contribution guidelines

keghani:


CONTRIBUTING.md
```diff
@@ -2,9 +2,49 @@
...
+For changes specific to windows builds, see [Filing a code
+issue](#filing-a-code-issue) and make sure you have found a reviewer and tester
+before sending the pull request.
+
+We are happy to review and accept pull requests for the following:
+
+*   Carrier updates
+*   Geocoding updates
+*   Bug fixes
+*   API changes
```
lararennie: What kinds of API changes? I'm not sure we want to.

CONTRIBUTING.md
```diff
@@ -2,9 +2,49 @@
...
+*   Documentation and code hygiene cleanups
+
+This list is not exhaustive.
+
+To get your pull request merged, we need the following:
+
+*   A CLA signature. See below.
+*   Responsiveness to back and forth review comments, because we will do a code
+    review.
+*   A clear description of the issue the pull request intends to solve.
```
lararennie: Should we require them to file a separate issue too?

CONTRIBUTING.md
```diff
@@ -2,9 +2,49 @@
...
+*   Tests that illustrate the issue, if applicable.
+*   Willingness and ability to implement the changes in C++, Java, and JS if
+    applicable. For example, changes to `PhoneNumberUtil.java` must be ported to
+    `phonenumberutil.cc` and `phonenumberutil.js`.
+    *   The initial pull request may contain the implementation in only one
+        language to get approval on the approach.
+
+If this is not possible, please file an issue.
```
lararennie: an issue "instead"

CONTRIBUTING.md
```diff
@@ -2,9 +2,49 @@
...
+
+Before spending significant time on a bug fix or API change pull request, we
+strongly encourage that you file an issue so we can discuss whether this is
+something we think should be fixed, how to do it, and whether we'd accept a pull
+request.
```
lararennie: Or for an API change, they could email the discussion list.

CONTRIBUTING.md
keghani: Yeah I removed from here. I just thought we had accepted such things recently?
Should I also exclude above? I think there may be cases where e.g. someone emails us asking if a function can be public and we think a variant of it could and ask for a PR.

CONTRIBUTING.md
keghani: Yes, added.

CONTRIBUTING.md
keghani: done

CONTRIBUTING.md
keghani: Why? To see if others see a need for the API?
As in why not an issue?

I think the bug fix thing could also be discussed in the list but it's an additional hurdle (one has to have a Google account and be whitelisted, so they're blocked until we do that).

keghani:
I address comments and re-ordered one thing.

PR #1853, opened 2017-07-14
Move repo-owned wiki pages to source tree and add quick links to main page + reorganize it

keghani:


keghani:
Also I added an intro page to the wiki: https://github.com/googlei18n/libphonenumber/wiki
Because apparently there needs to be a "Home" page. When I renamed the existing "Home" to "Android Studio setup" because that's what it was, I got an automatic new "Home" page.

Also I will delete these 2 after the PR is in:
https://github.com/googlei18n/libphonenumber/wiki/HowToMakeMetadataChanges
https://github.com/googlei18n/libphonenumber/wiki/HowToRunJavaDemo

keghani:
Want to draw your attention to one particular change: I'm proposing we remove the explicit version number from the demo heading.

All ready :) Sorry for the million iterations, kept finding things after it was in PR mode...

lararennie:
Someone filed a bug before about the demo thing and we added the v number to address that. I think we should either have it there or on the demo page itself and the latter is much harder to maintain.

README.md
```diff
@@ -94,22 +122,28 @@ try {
...
+
 ```json
 {
   "country_code": 41,
   "national_number": 446681800
 }
 ```
 
-` PhoneNumber ` is a class that is auto-generated from the phonenumber.proto with necessary modifications for efficiency. For details on the meaning of each field, refer to https://github.com/googlei18n/libphonenumber/blob/master/resources/phonenumber.proto
+`PhoneNumber` is a class that is auto-generated from `phonenumber.proto` with
```
lararennie: Is generated -> was originally generated

README.md
```diff
@@ -135,7 +171,8 @@ System.out.println(formatter.inputDigit('6'));  // Outputs "6"
 System.out.println(formatter.inputDigit('3'));  // Now outputs "650 253"
 ```
 
-### Geocoding Phone Numbers offline
+## Geocoding Phone Numbers offline
```
lararennie: Maybe we should just remove offline in the title. Our whole lib is offline.

README.md
```diff
@@ -146,7 +183,8 @@ System.out.println(geocoder.getDescriptionForNumber(swissNumberProto, Locale.GER
 System.out.println(geocoder.getDescriptionForNumber(swissNumberProto, Locale.ITALIAN));
 ```
 
-### Mapping Phone Numbers to carrier
+## Mapping Phone Numbers to carriers
```
lararennie: To original carriers? Maybe a caveat on portability is worthwhile

making_metadata_changes.md
```diff
@@ -0,0 +1,85 @@
...
+These steps outline how to edit the metadata for the phone number library to fix
+problems with validation or formatting.
+
+Note that we are currently not accepting validation or formatting pull requests,
+but you may use these guidelines to create changes in your own fork until we fix
+issues upstream.
+
+You can read more about the pull request [contribution
+guidelines](http://github.com/googlei18n/libphonenumber/blob/master/CONTRIBUTING.md#pull-requests)
+and file an issue if we should make validation fixes!
```
lararennie: Validation: metadata? Or validation or formatting?

making_metadata_changes.md
```diff
@@ -0,0 +1,85 @@
...
+*   `resources/ShortNumberMetadata.xml` for short-code data
+*   `resources/PhoneNumberAlternateFormats.xml` for alternate patterns for phone
+    number matching
+*   `resources/geocoding/xx` for geocoding data (where `xx` is the language code
+    you wish to edit)
+*   `resources/carrier/xx` for carrier mapping data
+
+Note:
+
+*   Include sources for data where appropriate
```
lararennie: This comment about sources we should move to the contribution guidelines for PRs?

keghani:
They can get it from the release page I linked to next to it?
It's more accurate than having it on the title. The 2 demos are never updated together (JS is from head, Java comes afterwards because head needs to be released to Maven first). The new wording says Java's going to be what's at the release page most of the time, with some lag.

We could also add the version number there explicitly but I don't see what that does other than manually have to update the number to what's at the release link. Still, could do it. I'd prefer to explicitly call it out as the Java version though, because JS is always what's at head.

README.md
```diff
@@ -9,41 +9,67 @@ validating international phone numbers. The Java version is optimized for
...
+*   Reporting an issue? Want to send a pull request? See the [contribution
+    guidelines](CONTRIBUTING.md)
+*   Check the [frequently asked questions](FAQ.md)
+*   Fun! [Falsehoods Programmers Believe About Phone Numbers](FALSEHOODS.md)
+*   Look for
+    [`README`s](http://github.com/googlei18n/libphonenumber/find/master) in
+    directories relevant to the code you're interested in.
+*   For porters:
+    *   [How to make metadata changes](making_metadata_changes.md)
+    *   [How to run the Java demo](run_java_demo.md)
```
keghani: Done

README.md
```diff
@@ -9,41 +9,67 @@ validating international phone numbers. The Java version is optimized for
...
-# Code
-To include the code in your application, either integrate with Maven or
-download the latest Jars from the Maven repository:
-
-http://repo1.maven.org/maven2/com/googlecode/libphonenumber/libphonenumber/
+*   Parsing / formatting / validating phone numbers for all countries / regions
+    of the world.
+*   `getNumberType` - gets the type of the number based on the number itself;
+    able to distinguish Fixed-line, Mobile, Toll-free, Premium Rate, Shared
+    Cost, VoIP and Personal Numbers  (whenever feasible).
```
keghani: Done

README.md
```diff
@@ -9,41 +9,67 @@ validating international phone numbers. The Java version is optimized for
...
+*   `isNumberMatch` - gets a confidence level on whether two numbers could be
+    the same.
+*   `getExampleNumber` / `getExampleNumberByType` - provides valid example
+    numbers for all countries/regions, with the option of specifying which type
+    of example phone number is needed.
+*   `isPossibleNumber` - quickly guessing whether a number is a possible
+    phone number by using only the length information, much faster than a full
```
keghani: Done

README.md
```diff
@@ -9,41 +9,67 @@ validating international phone numbers. The Java version is optimized for
...
+    validation.
+*   `isValidNumber` - full validation of a phone number for a region using
+    length and prefix information.
+*   `AsYouTypeFormatter` - formats phone numbers on-the-fly when users enter
+    each digit.
+*   `findNumbers` - finds numbers in text input.
```
keghani: Done

README.md
```diff
@@ -9,41 +9,67 @@ validating international phone numbers. The Java version is optimized for
...
+*   `PhoneNumberOfflineGeocoder` - provides geographical information related to
+    a phone number.
+*   `PhoneNumberToCarrierMapper` - provides carrier information related to a
+    phone number.
+*   `PhoneNumberToTimeZonesMapper` - provides timezone information related to a
+    phone number.
+
+# Demo
+
+The [Java demo](http://libphonenumber.appspot.com/) is updated periodically,
```
keghani: Changed from:

```
The [Java demo](http://libphonenumber.appspot.com/) is updated periodically,
with a slight delay after the GitHub release. See the [latest release's version
number](http://github.com/googlei18n/libphonenumber/releases).
```

to:
```
The [Java demo](http://libphonenumber.appspot.com/) is updated with a slight
delay after the GitHub release.
Last demo update: v8.6.0.
If this number is lower than the [latest release's version
number](http://github.com/googlei18n/libphonenumber/releases), we are between
releases and the demo may be at either version.
```

To answer the other comment, I don't think the number on the demo should come from the GitHub page, because that's incorrect too. Between the GitHub release and the demo update it would say the new number but not be updated yet.

I still don't see why anyone would want the number on the README page when it's incorrect some of the time, and they need to know the caveat about it being off from the release. I think the only potential improvement over not having anything there would be to explicitly say the demo **was** updated, so I put that other language in.

README.md
keghani: Done

README.md
keghani: Done

README.md
keghani: Done.

New language:

```
Caveat: We do not provide data about the current carrier of a phone number, only
the original carrier who is assigned the corresponding range. Read about [number
portability](FAQ.md#what-is-mobile-number-portability).
```

making_metadata_changes.md
keghani: Updated whole section.

making_metadata_changes.md
keghani: I don't think it should be there - this was from the existing wiki page, and applies only to PRs of metadata changes in xml files, which we don't do PRs for now.

I re-worded. I want to keep the idea of it in this doc, in case we bring back xml PRs, this point should be there.

keghani:
Ready for another look.

README.md
```diff
@@ -9,41 +9,75 @@ validating international phone numbers. The Java version is optimized for
...
-[Java](http://libphonenumber.appspot.com/)
-
-[JavaScript](https://rawgit.com/googlei18n/libphonenumber/master/javascript/i18n/phonenumbers/demo-compiled.html)
-
-# Code
-To include the code in your application, either integrate with Maven or
-download the latest Jars from the Maven repository:
-
-http://repo1.maven.org/maven2/com/googlecode/libphonenumber/libphonenumber/
+*   Parsing / formatting / validating phone numbers for all countries / regions
```
lararennie: Grammarly.com backs me up here :)
"Many people wonder whether to use a space before and after slashes in a sentence. The answer is typically no on both fronts. Spaces before slashes should be avoided. "

Slashes can be followed by space if it is to break up several short-lines in a poem etc.

keghani: Done

PR #1894, opened 2017-08-07
Improve parsing logic to be smarter about national-prefix detection & stripping.

andyst:
Make use of the recently introduced validation enum values IS_POSSIBLE_LOCAL_ONLY and INVALID_LENGTH.
Helps with e.g. short-code parsing where the short-code's prefix matches the nationalPrefix.

E.g. Iran has nationalPrefix=0 and short-codes 096\d{2} should be added. To ensure the "0" isn't stripped (but preserved as italian_leading_zero), we need to make sure that the parser doesn't consider the 0 as the national-prefix in this number.

This already worked if:
- nationalPrefix != prefix of short-code, or
- nationalPrefix == prefix of short code and:
  - short-code matches generalDesc pattern but remainder doesn't, or
  - remainder of short-code is too short to be a non-short-code.

The last bullet item is why 112 already works as US short-code although nationalPrefix=1.

The fix is necessary for the case where the remainder (number without the national prefix) isn't too short but is possible local-only.

### DIFF
--- a/javascript/i18n/phonenumbers/phonenumberutil.js
+++ b/javascript/i18n/phonenumbers/phonenumberutil.js
@@ -4215,9 +4215,14 @@ i18n.phonenumbers.PhoneNumberUtil.prototype.parseHelper_ =
     // carrier code be long enough to be a possible length for the region.
     // Otherwise, we don't do the stripping, since the original number could be
     // a valid short number.
-    if (this.testNumberLength_(potentialNationalNumber.toString(),
-            regionMetadata) !=
-        i18n.phonenumbers.PhoneNumberUtil.ValidationResult.TOO_SHORT) {
+    var validatonResult = this.testNumberLength_(
+        potentialNationalNumber.toString(), regionMetadata);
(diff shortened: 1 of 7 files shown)


javascript/i18n/phonenumbers/phonenumberutil.js
```diff
@@ -4215,9 +4215,14 @@ i18n.phonenumbers.PhoneNumberUtil.prototype.parseHelper_ =
     // carrier code be long enough to be a possible length for the region.
     // Otherwise, we don't do the stripping, since the original number could be
     // a valid short number.
-    if (this.testNumberLength_(potentialNationalNumber.toString(),
-            regionMetadata) !=
-        i18n.phonenumbers.PhoneNumberUtil.ValidationResult.TOO_SHORT) {
+    var validatonResult = this.testNumberLength_(
```
keghani: validationResult (with a second i)

andyst:
Thanks Keghani!

PTAL

PR #1943, opened 2017-09-27
Support dialer commands in extension

voronaam:
Continuing the work on a fix for https://github.com/googlei18n/libphonenumber/issues/467

This PR only contains the changes for Java and is likely to be rejected. But if Java portion of it is approved, I will make the same changes to the other languages libraries in the project.

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -305,8 +305,8 @@ public class PhoneNumberUtil {
   private static final String DEFAULT_EXTN_PREFIX = " ext. ";
 
   // Pattern to capture digits used in an extension. Places a maximum length of "7" for an
-  // extension.
-  private static final String CAPTURING_EXTN_DIGITS = "(" + DIGITS + "{1,7})";
+  // extension. Commas are allowed, as long they are not the trailing symbol
+  private static final String CAPTURING_EXTN_DIGITS = "([\\p{Nd},]{0,6}\\p{Nd})";
   // Regexp of all possible ways to write extensions, for use when parsing. This will be run as a
   // case-insensitive regexp match. Wide character versions are also provided after each ASCII
   // version.
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -305,8 +305,14 @@ public class PhoneNumberUtil {
   private static final String DEFAULT_EXTN_PREFIX = " ext. ";
 
   // Pattern to capture digits used in an extension. Places a maximum length of "7" for an
-  // extension. Commas are allowed, as long they are not the trailing symbol
-  private static final String CAPTURING_EXTN_DIGITS = "([\\p{Nd},]{0,6}\\p{Nd})";
+  // extension.
+  private static final String CAPTURING_EXTN_DIGITS = "(\\p{Nd}{1,7})";
+
+  // Pattern to capture digits and commands used in an extension. Places a maximum length of "7"
+  // for an entire extension. Comma and semicolon are allowed, as long as they are not the trailing symbol.
+  // They denote dialing commands to delay (comma) or wait for a tone (semicolon).
+  private static final String CAPTURING_EXTN_DIGITS_AND_COMMANDS = "([\\p{Nd},;]{0,6}\\p{Nd})";
+
   // Regexp of all possible ways to write extensions, for use when parsing. This will be run as a
   // case-insensitive regexp match. Wide character versions are also provided after each ASCII
   // version.
@@ -320,15 +326,16 @@ public class PhoneNumberUtil {
     // hardly ever used to indicate this.
     String singleExtnSymbolsForParsing = ",;" + singleExtnSymbolsForMatching;
 
-    EXTN_PATTERNS_FOR_PARSING = createExtnPattern(singleExtnSymbolsForParsing);
-    EXTN_PATTERNS_FOR_MATCHING = createExtnPattern(singleExtnSymbolsForMatching);
+    EXTN_PATTERNS_FOR_PARSING = createExtnPattern(singleExtnSymbolsForParsing, CAPTURING_EXTN_DIGITS_AND_COMMANDS);
+    EXTN_PATTERNS_FOR_MATCHING = createExtnPattern(singleExtnSymbolsForMatching, CAPTURING_EXTN_DIGITS);
   }
 
   /**
    * Helper initialiser method to create the regular-expression pattern to match extensions,
    * allowing the one-char extension symbols provided by {@code singleExtnSymbols}.
+   * {@code extensionPattern} allows to specify the extension capturing pattern.
    */
-  private static String createExtnPattern(String singleExtnSymbols) {
+  private static String createExtnPattern(String singleExtnSymbols, String extensionPattern) {
     // There are three regular expressions here. The first covers RFC 3966 format, where the
     // extension is added using ";ext=". The second more generic one starts with optional white
     // space and ends with an optional full stop (.), followed by zero or more spaces/tabs/commas
@@ -339,10 +346,10 @@ public class PhoneNumberUtil {
     // Canonical-equivalence doesn't seem to be an option with Android java, so we allow two options
     // for representing the accented o - the character itself, and one in the unicode decomposed
     // form with the combining acute accent.
-    return (RFC3966_EXTN_PREFIX + CAPTURING_EXTN_DIGITS + "|" + "[ \u00A0\\t,]*"
+    return (RFC3966_EXTN_PREFIX + extensionPattern + "|" + "[ \u00A0\\t,]*"
         + "(?:e?xt(?:ensi(?:o\u0301?|\u00F3))?n?|\uFF45?\uFF58\uFF54\uFF4E?|"
         + "[" + singleExtnSymbols + "]|int|anexo|\uFF49\uFF4E\uFF54)"
-        + "[:\\.\uFF0E]?[ \u00A0\\t,-]*" + CAPTURING_EXTN_DIGITS + "#?|"
+        + "[:\\.\uFF0E]?[ \u00A0\\t,-]*" + extensionPattern + "#?|"
         + "[- ]+(" + DIGITS + "{1,5})#");
   }
 
@@ -741,6 +748,7 @@ public class PhoneNumberUtil {
   /**
    * Normalizes a string of characters representing a phone number. This strips all characters which
    * are not diallable on a mobile phone keypad (including all non-ASCII digits).
+   * Also strips out possible dialer commands from the extension (comma and seimcolon).
    *
    * @param number  a string of characters representing a phone number
    * @return  the normalized string version of the phone number
@@ -3337,7 +3345,7 @@ public class PhoneNumberUtil {
     PhoneNumber secondNumber = copyCoreFieldsOnly(secondNumberIn);
     // Early exit if both had extensions and these are different.
     if (firstNumber.hasExtension() && secondNumber.hasExtension()
-        && !firstNumber.getExtension().equals(secondNumber.getExtension())) {
+        && !extensionMatches(firstNumber.getExtension(), secondNumber.getExtension())) {
       return MatchType.NO_MATCH;
     }
     int firstNumberCountryCode = firstNumber.getCountryCode();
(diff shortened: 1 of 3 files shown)


mizuni:
Hi, sorry for the delay, but we have some folks out at the moment, and it will take more time for us to regroup to review and decide on the best next steps.

mizuni:
In the mean time, would you also be able to provide some example use cases and an idea of the impact? Thanks!

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberMatcherTest.java
```diff
@@ -423,6 +423,42 @@ public void testDoesNotMatchMultiplePhoneNumbersSeparatedWithNoWhiteSpace() thro
...
+    assertEquals(match2, matches.next());
+  }
+
+  public void testMatchesPhoneNumberWithExtensionWithExtraPunctuation() throws Exception {
+    String text = "Call 650-253-4561x25,455";
+    String region = RegionCode.US;
+
+    PhoneNumber number1 = new PhoneNumber();
+    number1.setCountryCode(phoneUtil.getCountryCodeForRegion(region));
+    number1.setNationalNumber(6502534561L);
```
GoogleJacob: 8642014971

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberMatcherTest.java
```diff
@@ -423,6 +423,42 @@ public void testDoesNotMatchMultiplePhoneNumbersSeparatedWithNoWhiteSpace() thro
...
+    assertEquals(match1, matches.next());
+    assertEquals(match2, matches.next());
+  }
+
+  public void testMatchesPhoneNumberWithExtensionWithExtraPunctuation() throws Exception {
+    String text = "Call 650-253-4561x25,455";
+    String region = RegionCode.US;
+
+    PhoneNumber number1 = new PhoneNumber();
+    number1.setCountryCode(phoneUtil.getCountryCodeForRegion(region));
```
GoogleJacob: 8642014971

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberMatcherTest.java
andyst: @GoogleJacob first time I see you comment on this project. What's your interest in / relationship with this change / pull-request?

Also, regarding this comment and the other comment, what do you mean? It's a bit cryptic. Please elaborate.

andyst:
@voronaam can you clarify / summarize what has changed since lararennie reviewed and commented on the proposed change (since https://github.com/googlei18n/libphonenumber/issues/467#issuecomment-299289575 and https://github.com/googlei18n/libphonenumber/issues/467#issuecomment-315312691)?

Thanks!

voronaam:
@andyst I have addressed the concern around the `findNumbers` functionality. Separated the regular expression used in it, such that my changes to the extension dial commands only affect phone number parsing.

I think I have tests covering the changes.

It was agreed previously, that once Java version is Ok'ed, I would make the same changes in C and JS versions of the library.

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -306,7 +306,13 @@
 
   // Pattern to capture digits used in an extension. Places a maximum length of "7" for an
   // extension.
-  private static final String CAPTURING_EXTN_DIGITS = "(" + DIGITS + "{1,7})";
+  private static final String CAPTURING_EXTN_DIGITS = "(\\p{Nd}{1,7})";
```
andyst: Please revert this back to reusing DIGITS rather than writing "\\p{Nd}"

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -306,7 +306,13 @@
...
+
+  // Pattern to capture digits and commands used in an extension. Places a maximum length of "7"
+  // for an entire extension. Comma and semicolon are allowed, as long as they are not the trailing symbol.
+  // They denote dialing commands to delay (comma) or wait for a tone (semicolon).
+  private static final String CAPTURING_EXTN_DIGITS_AND_COMMANDS = "([\\p{Nd},;]{0,6}\\p{Nd})";
```
andyst: Please use DIGITS instead of writing "\\p{Nd}"

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -306,7 +306,13 @@
 
   // Pattern to capture digits used in an extension. Places a maximum length of "7" for an
   // extension.
-  private static final String CAPTURING_EXTN_DIGITS = "(" + DIGITS + "{1,7})";
+  private static final String CAPTURING_EXTN_DIGITS = "(\\p{Nd}{1,7})";
+
+  // Pattern to capture digits and commands used in an extension. Places a maximum length of "7"
+  // for an entire extension. Comma and semicolon are allowed, as long as they are not the trailing symbol.
+  // They denote dialing commands to delay (comma) or wait for a tone (semicolon).
```
andyst: Please reference the standards (URL?) which mention comma and semicolon for this purpose.

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -320,15 +326,16 @@
...
+    EXTN_PATTERNS_FOR_MATCHING = createExtnPattern(singleExtnSymbolsForMatching, CAPTURING_EXTN_DIGITS);
   }
 
   /**
    * Helper initialiser method to create the regular-expression pattern to match extensions,
    * allowing the one-char extension symbols provided by {@code singleExtnSymbols}.
+   * {@code extensionPattern} allows to specify the extension capturing pattern.
    */
-  private static String createExtnPattern(String singleExtnSymbols) {
+  private static String createExtnPattern(String singleExtnSymbols, String extensionPattern) {
```
andyst: Instead of extensionPattern I'd call it capturingExtnDigits because the comment below explains that there are three regular expressions involved and arguably those are "extension patterns" too (e.g. RFC3966_EXTN_PREFIX).

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -741,6 +748,7 @@ static StringBuilder normalizeDigits(CharSequence number, boolean keepNonDigits)
   /**
    * Normalizes a string of characters representing a phone number. This strips all characters which
    * are not diallable on a mobile phone keypad (including all non-ASCII digits).
+   * Also strips out possible dialer commands from the extension (comma and seimcolon).
```
andyst: typo seimcolon -> semicolon

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberMatcherTest.java
andyst: @voronaam: Please ignore GoogleJacob's comments (unless we hear back).

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -3369,6 +3377,16 @@ public MatchType isNumberMatch(PhoneNumber firstNumberIn, PhoneNumber secondNumb
     return MatchType.NO_MATCH;
   }
 
+  // Return true if extension matches, possibly ignoring control characters
```
andyst: Please add sentence-ending period.

andyst: I'd remove "possibly" because we are ignoring them (whether or not they're present).

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -3369,6 +3377,16 @@ public MatchType isNumberMatch(PhoneNumber firstNumberIn, PhoneNumber secondNumb
...
+  private boolean extensionMatches(String firstExtension, String secondExtension) {
+      if (firstExtension.equals(secondExtension)) {
+          return true;
+      }
+      String firstSanitized = firstExtension.replaceAll("[,;]", "");
```
andyst: Performance: Each time we compare 2 phone numbers which have extensions, we'd hit this code where uncompiled regular expressions are used.

I think we should go with:
String firstSanitized = firstExtension.replace(',', '').replace(';', '');

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -3337,7 +3345,7 @@ public MatchType isNumberMatch(PhoneNumber firstNumberIn, PhoneNumber secondNumb
     PhoneNumber secondNumber = copyCoreFieldsOnly(secondNumberIn);
     // Early exit if both had extensions and these are different.
     if (firstNumber.hasExtension() && secondNumber.hasExtension()
-        && !firstNumber.getExtension().equals(secondNumber.getExtension())) {
+        && !extensionMatches(firstNumber.getExtension(), secondNumber.getExtension())) {
```
andyst: The Javadoc above should mention what happens with differences in control-characters (comma & semicolon) in the extensions.

Before addressing this, let's figure out the semantics. For that, please see my comment on testIsNumberMatchMatches.

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberUtilTest.java
```diff
@@ -2652,6 +2660,14 @@ public void testParseExtensions() throws Exception {
     assertEquals(usWithExtension, phoneUtil.parse("+1 (645) 123 1234-910#", RegionCode.US));
     // Retry with the same number in a slightly different format.
     assertEquals(usWithExtension, phoneUtil.parse("+1 (645) 123 1234 ext. 910#", RegionCode.US));
+
+    // Comma and semicolon are allowed in extensions
+    usWithExtension.clear();
+    usWithExtension.setCountryCode(1).setNationalNumber(6451231234L).setExtension("91,0");
+    assertEquals(usWithExtension, phoneUtil.parse("+1 (645) 123 1234 ext: 91,0", RegionCode.US));
+    usWithExtension.clear();
+    usWithExtension.setCountryCode(1).setNationalNumber(6451231234L).setExtension("91;0");
```
andyst: Please exercise the allowed pattern a bit more.
(both allowed and disallowed)

- test which shows what happens where there's a trailing comma or semicolon.
- multiple control characters in a row (but not at the end)
- extension starting with a control character
- "92-3" and "91.0" i.e. what happens if it's another punctuation character appears in the extension

java/libphonenumber/test/com/google/i18n/phonenumbers/PhoneNumberUtilTest.java
```diff
@@ -2770,6 +2786,11 @@ public void testIsNumberMatchMatches() throws Exception {
                  phoneUtil.isNumberMatch("+64 3 331-6005 extn 1234", "+6433316005#1234"));
     assertEquals(PhoneNumberUtil.MatchType.EXACT_MATCH,
                  phoneUtil.isNumberMatch("+64 3 331-6005 ext. 1234", "+6433316005;1234"));
+    // Test numbers with extension with modem control characters in them
+    assertEquals(PhoneNumberUtil.MatchType.SHORT_NSN_MATCH,
```
andyst: SHORT_NSN_MATCH doesn't seem quite right, but maybe it is the best here.

We could return EXACT_MATCH if we specify that such differences are ignored. But most users of EXACT_MATCH may depend on exact equivalence of the PhoneNumber object / formatted result.

It's not worth introducing a new enum value for differences in the commands in the extension.

Another alternative would be if the parse(...) method drops the command characters. I.e. the extension as stored in the resulting PhoneNumber object would never include command characters.

1. What do you think of that option? I probably haven't thought this through.

2. SHORT_NSN_MATCH seems reasonable. What do you think?

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
edent: This is part of ITU-T V.250 https://www.itu.int/rec/T-REC-V.250-200307-I/en

Specifically 6.3.1.2 Pause during dialling, and 5.4.5.1 Concatenating commands after extended syntax commands, from https://www.itu.int/rec/dologin_pub.asp?lang=e&id=T-REC-V.250-200307-I!!PDF-E&type=items

voronaam:
Deprecating this PR for no movement on it.

PR #1946, opened 2017-10-03
Update C# port link

twcclegg:
The ownership of the C# port has changed, you will note that the old location https://github.com/aidanbebbington/libphonenumber-csharp/ redirects to the new location.


twcclegg:
I signed it!
Not sure if it's needed in this case.

keghani:
Thanks! Approving...

PR #1955, opened 2017-10-09
Move from readdir_r to readdir

### LINKED ISSUE #1307: Use readdir instead of readdir_r
I use an Arch Linux system with gcc 6.1.1 and glibc 2.24. When building the C++ version of the library I get the following error:

```
libphonenumber/tools/cpp/src/cpp-build/generate_geocoding_data.cc: In Funktion »bool i18n::phonenumbers::ListDirectory(const string&, std::vector<i18n::phonenumbers::DirEntry>*)«:
libphonenumber/tools/cpp/src/cpp-build/generate_geocoding_data.cc:108:21: Fehler: »int readdir_r(DIR*, dirent*, dirent**)« ist veraltet [-Werror=deprecated-declarations]
     const int res = readdir_r(dir, &entry, &dir_result);
                     ^~~~~~~~~
externals/libphonenumber/tools/cpp/src/cpp-build/generate_geocoding_data.cc:19:0:
/usr/include/dirent.h:183:12: Anmerkung: hier deklariert
 extern int readdir_r (DIR *__restrict __dirp,
            ^~~~~~~~~
libphonenumber/tools/cpp/src/cpp-build/generate_geocoding_data.cc:108:55: Fehler: »int readdir_r(DIR*, dirent*, dirent**)« ist veraltet [-Werror=deprecated-declarations]
     const int res = readdir_r(dir, &entry, &dir_result);
                                                       ^
libphonenumber/tools/cpp/src/cpp-build/generate_geocoding_data.cc:19:0:
/usr/include/dirent.h:183:12: Anmerkung: hier deklariert
 extern int readdir_r (DIR *__restrict __dirp,
```

readdir_r is deprecated.

keghani:
b/31256592
Fixes #1307

Relevant man pages:
https://linux.die.net/man/3/readdir
https://linux.die.net/man/3/errno

### DIFF
--- a/tools/cpp/src/cpp-build/generate_geocoding_data.cc
+++ b/tools/cpp/src/cpp-build/generate_geocoding_data.cc
@@ -105,11 +105,13 @@ bool ListDirectory(const string& path, vector<DirEntry>* entries) {
   struct dirent *entry;
   struct stat entry_stat;
   while (true) {
+    int errno_last = errno;
     errno = 0;
     entry = readdir(dir);
     if (errno) {
       return false;
     }
+    errno = errno_last;
     if (entry == NULL) {
       return true;
     }
--- a/tools/cpp/src/cpp-build/generate_geocoding_data.cc
+++ b/tools/cpp/src/cpp-build/generate_geocoding_data.cc
@@ -106,6 +106,10 @@ bool ListDirectory(const string& path, vector<DirEntry>* entries) {
   struct dirent *entry;
   struct stat entry_stat;
   while (true) {
+    // Set errno to 0 to be able to check if an error occurs during the
+    // readdir() call. NULL is the return value when the end of the directory
+    // stream is reached or when an error occurs, and the errno check is the
+    // only thing that helps us distinguish between the two cases.
     errno = 0;
     entry = readdir(dir);
     if (entry == NULL) {
--- a/tools/cpp/src/cpp-build/generate_geocoding_data.cc
+++ b/tools/cpp/src/cpp-build/generate_geocoding_data.cc
@@ -17,6 +17,7 @@
 #include "cpp-build/generate_geocoding_data.h"
 
 #include <dirent.h>
+#include <errno.h>
 #include <locale>
 #include <sys/stat.h>
 #include <algorithm>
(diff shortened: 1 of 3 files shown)


tools/cpp/src/cpp-build/generate_geocoding_data.cc
```diff
@@ -102,20 +102,21 @@ bool ListDirectory(const string& path, vector<DirEntry>* entries) {
...
   AutoCloser<DIR> dir_closer(&dir, closedir);
-  struct dirent entry, *dir_result;
+  struct dirent *entry;
   struct stat entry_stat;
   while (true) {
-    const int res = readdir_r(dir, &entry, &dir_result);
-    if (res) {
+    errno = 0;
+    entry = readdir(dir);
+    if (errno) {
```
roubert: The standard does actually not guarantee that a call to readdir() that returns a non-NULL pointer doesn't also modify errno, that guarantee is only specified for a call that returns NULL. So checking errno before checking the return value could possibly yield a false positive.

A safer implementation would be:

if (entry == NULL) {
  return errno == 0;
}

keghani:
Thanks Fredrik!
I had interpreted from https://linux.die.net/man/3/errno that if my call doesn't return -1 for success, I couldn't just leave errno at 0:

> Valid error numbers are all nonzero; errno is never set to zero by any system call or library function.

> For some system calls and library functions (e.g., getpriority(2)), -1 is a valid return on success. In such cases, a successful return can be distinguished from an error return by setting errno to zero before the call, and then, if the call returns a status that indicates that an error may have occurred, checking to see if errno has a nonzero value.

But I trust your judgment.
Made the changes. Thanks a lot!

tools/cpp/src/cpp-build/generate_geocoding_data.cc
```diff
@@ -102,20 +103,18 @@ bool ListDirectory(const string& path, vector<DirEntry>* entries) {
...
+  struct dirent *entry;
   struct stat entry_stat;
   while (true) {
-    const int res = readdir_r(dir, &entry, &dir_result);
-    if (res) {
-      return false;
-    }
-    if (dir_result == NULL) {
-      return true;
+    errno = 0;
```
hagbard: Given Fredrik's comment that this is slightly unobvious behaviour, can you add a comment explaining this here.
Setting errno explicitly to success before a call is not something I'm used to seeing.

keghani: Fredrik's comment was on an outdated version of the file, I changed it. This is what he had suggested and if IIUC he doesn't think this one's unobvious - what he didn't think I should do was save the previous value of errno and set back to it if I come back with 0, which is what I had at commit 53642bf

Maybe let's discuss this in person. I'm happy with implementing it any way since you both know more here, but I'd like to understand it.

keghani:
No worries, I read up some more and found http://pubs.opengroup.org/onlinepubs/009695399/functions/readdir.html and my questions are answered. This is the recommended way to use readdir, and I did add a comment to explain.

keghani:
@roubert Done, thanks!
@hagbard ready for final look

PR #1961, opened 2017-10-11
Add link to android-optimized port

ChaitanyaPramod:
libphonenumber-android repackages the library and replaces nonoptimal usage of `Class#getResourceAsStream()` with `AssetManager#open()` for android as suggested in [FAQ](https://github.com/googlei18n/libphonenumber/blob/e95e140dd7eeb898b9ebe1eee98ef63048ace1c0/FAQ.md#optimize-loads)

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,11 @@ However, we emphasize that these ports are by developers outside the
 libphonenumber project. We do not evaluate their quality or influence their
 maintenance processes.
 
+*   Android: Java version of libphonenumber can be used on Android as is.
+    https://github.com/MichaelRocks/libphonenumber-android is a repackaged
+    port optimized for Android by using `AssetManager#open()` instead of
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -229,3 +229,4 @@ maintenance processes.
 *   PostgreSQL in-database types: https://github.com/blm768/pg-libphonenumber
 *   Python: https://github.com/daviddrysdale/python-phonenumbers
 *   Ruby: https://github.com/mobi/telephone_number
+*   Android: https://github.com/MichaelRocks/libphonenumber-android
```
keghani: 1. I would say Android-optimized (since our port is actually also used in Android.

2. It would be nice to include the info you had in the PR description about assets as well. Something like the wording used to describe the other Javascript versions.

3. Please put this first to keep the alpha sorting

ChaitanyaPramod: Updated the PR with the comments. Calling it Android so it's similar to JS ports entry.

keghani:
Thanks for contributing!

PR #2021, opened 2017-11-17
Update Javadoc for PhoneNumberUtil.PhoneNumberFormat enum reg E.123 format rules we follow

penmetsaa:
We say that "INTERNATIONAL and NATIONAL formats are consistent with the definition in ITU-T Recommendation E.123." This is not completely true as we follow local conventions which can include using '-' instead of whitespace (Mentioned in E.123 doc) as separators. PTAL at attached bug for more details.

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -389,7 +389,7 @@ public class PhoneNumberUtil {
    * will have a prefix of "tel:" added, e.g. "tel:+41-44-668-1800".
    *
    * Note: If you are considering storing the number in a neutral format, you are highly advised to
-   * use the phonenumber.proto. For details, see http://go/phone-pb
+   * use the PhoneNumber class.
    */
   public enum PhoneNumberFormat {
     E164,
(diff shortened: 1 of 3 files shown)


java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
```diff
@@ -380,15 +380,16 @@ private static String createExtnPattern(String singleExtnSymbols) {
...
+   * separators. For example, the number of the Google Switzerland office will be written as
+   * "+41 44 668 1800" in INTERNATIONAL format, and as "044 668 1800" in NATIONAL format. E164
+   * format is as per INTERNATIONAL format but with no formatting applied, e.g. "+41446681800".
+   * RFC3966 is as per INTERNATIONAL format, but with all spaces and other separating symbols
+   * replaced with a hyphen, and with any phone number extension appended with ";ext=". It also
+   * will have a prefix of "tel:" added, e.g. "tel:+41-44-668-1800".
    *
    * Note: If you are considering storing the number in a neutral format, you are highly advised to
-   * use the PhoneNumber class.
+   * use the phonenumber.proto. For details, see http://go/phone-pb
```
andyst: Please revert the change on this line. Github uses a Java class, not a proto.

penmetsaa: Oh Sorry. My bad. Reverted this change now. PTAL.

andyst:
Thanks!

PR #2107, opened 2018-02-01
Fix #2002: port findNumbers() from Java to JavaScript

### LINKED ISSUE #2002: Port findNumbers() call to Javascript
This appears to still be missing, and though it's on the todo list in the readme I don't see an issue for this.

My need is for use in an email parser in a node.js app.

humphd:
This PR adds support for `findNumbers()`.  I’ve used the Java code as a guide, and tried to stick as closely to it as possible.  As such, there’s lots of things that could be made to use more common JS patterns; however, I wanted to get feedback before doing any of that, as I get the sense that uniformity across your 3 languages is more useful for maintenance.  Please let me know how I can fix things to adhere to your standards, as it wasn’t always obvious to me without existing patterns.

I also ported the tests, and so far I have 147 of them passing, with 35+ still failing.  I’m hoping that some of the issues will disappear after I fix any issues in the code a proper review uncovers.  I’ve commented out failing tests, and marked them with `// XXX_FAILING: …` in the source of the test files.  For example, all tests referencing the `ZZ` region fail, as the JS code doesn’t think it’s valid.  Other failures could easily be issues in my code for sure.

This is my first time using Closure, and I have what appears to be a circular reference problem between `i18n.phonenumbers.PhoneNumberMatcher` and `i18n.phonenumbers.PhoneNumberUtil`.  Likely I need to move some of the constants off of `i18n.phonenumbers.PhoneNumberUtil` and into `i18n.phonenumbers.PhoneNumberMatcher`, and use instance methods on `this.phoneUtil` within the PhoneNumberMatcher methods.  However, there might be something else I’m missing.

I’m a bit unclear on how to port the code block that uses `MetadataManager.getAlternateFormatsForCountry()` in `PhoneNumberMatcher.checkNumberGroupingIsValid()`.  It seems to me like some of this work might already be done in the `metadata.js` and `phonemetadata.pb.js` files, but I’m not following things 100%.  Some guidance on this would be helpful.

I’ve left other questions in the source marked with `// XXX: …` to help remind myself what I wanted to discuss with a reviewer.

Thanks for considering this.  I think having `findNumbers()` will make this a lot more useful for web and node developers.

### DIFF
--- a/javascript/i18n/phonenumbers/phonenumbermatch.js
+++ b/javascript/i18n/phonenumbers/phonenumbermatch.js
@@ -86,3 +86,15 @@ i18n.phonenumbers.PhoneNumberMatch.prototype.toString = function() {
         && number.equals(other.number);
   }
 **/
+
+i18n.phonenumbers.PhoneNumberMatch.prototype.equals = function(obj) {
+  if(this === obj) {
+    return true;
+  }
+  if(!(obj instanceof i18n.phonenumbers.PhoneNumberMatch)) {
+    return false;
+  }
+  return this.rawString == obj.rawString &&
+         this.start == obj.start         &&
+         this.number.equals(obj.number);
+};
--- a/javascript/i18n/phonenumbers/phonenumbermatch.js
+++ b/javascript/i18n/phonenumbers/phonenumbermatch.js
@@ -1,0 +1,88 @@
+/*
+ * Copyright (C) 2011 The Libphonenumber Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+goog.provide('i18n.phonenumbers.PhoneNumberMatch');
+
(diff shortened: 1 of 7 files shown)


keghani:
Hi @humphd,

Thanks for your contribution!

Please see the [contribution guidelines for pull requests](https://github.com/googlei18n/libphonenumber/blob/master/CONTRIBUTING.md#pull-requests).

We need to think in our team about whether this is something we want to fix right now given other priorities. I would not make any more changes to avoid spending cycles in case we can't accept it.

But in the meantime, it would be great if you could link to an issue as described at the link. There are also more details at the link as to what's required of a pull request to be merged, so to set expectations, e.g. it may take a while to do a thorough review and several iterations.

Next action: I'll get back to you on this thread.

keghani:
Ignore my request about an issue, it's already at https://github.com/googlei18n/libphonenumber/issues/2002

andyst:
@humphd thanks for the contribution! Unfortunately we don't have the bandwidth right now to provide in-depth guidance for a large code change like this at this point. It's easier to prioritize work which is aligned with primary use-cases of the library and so far there has been little demand for this functionality in JavaScript.

If you need this functionality now, I'd suggest to arrange it (for now) in a way to keep the functionality in separate files rather than touching `phonenumberutil.js` (i.e. moving the new bits related to matching which are in `phonenumberutil.js` over to `phonenumbermatcher.js`. That way you can have this functionality in a Github fork / your own code base without needing to maintain patches to libphonenumber.

We'll reevaluate the situation in a few months. Thanks!

humphd:
This is unfortunate, and somewhat surprising, given that https://github.com/googlei18n/libphonenumber/tree/master/javascript#todo lists this as needing to be done, and #2002 was left open, implying that a PR to fix it would be accepted.  If neither is the case, you should signal that in your docs/bugs.

PR #2110, opened 2018-02-05
Update FALSEHOODS.md

ctholho:
Update: Each country has only one country calling code
Seeing Monaco here is a surprise because it is a country 1500 to the west. I expected that it was a mistake and it should have been Kosovos neighbor, Macedonia. But no, Monaco is right. This should be highlighted.


ctholho:
I signed it!

PR #2117, opened 2018-02-07
Updated Mongolian carrier prefixes

digz6666:
Updated from Communications Regulatory Commission of Mongolian.
http://www.crc.gov.mn/k/4L/36

Another source: http://www.wtng.info/wtng-976-mn.html


digz6666:
I signed it!

PR #2131, opened 2018-02-15
adding deprecated forms of methods removed in 7e6feae

yinzara:
While String is a class that implements the CharSequence interface, changing method parameters from String to CharSequence on a public API causes runtime errors if a library is compiled with a prior version of libphonenumber (i.e. libphonenumber 8.6.0-) but runs with a more recent version of the library (8.7.0+).

This pull request adds back in methods with the original signatures that were removed/changed in commit 7e6feae but marks them as deprecated allowing the new versions of the library to be compatible with code compiled against the old versions of the library.


yinzara:
I signed the CLA before submitting this commit and pull request.

PR #2132, opened 2018-02-15
adding deprecated forms of methods removed in 7e6feae

yinzara:
While String is a class that implements the CharSequence interface, changing method parameters from String to CharSequence on a public API causes runtime errors if a library is compiled with a prior version of libphonenumber (i.e. libphonenumber 8.6.0-) but runs with a more recent version of the library (8.7.0+).

This pull request adds back in methods with the original signatures that were removed/changed in commit 7e6feae allowing the new versions of the library to be compatible with code compiled against the old versions of the library.

### DIFF
--- a/CONTRIBUTORS
+++ b/CONTRIBUTORS
@@ -52,6 +52,7 @@ Fredrik Roubert
 George Yakovlev
 Nico Weber
 Yoshifumi Inoue
+Matt Morrissette
 
 # Debian:
 Fredrik Roubert
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -724,6 +724,13 @@ public class PhoneNumberUtil {
   public static String normalizeDigitsOnly(CharSequence number) {
     return normalizeDigits(number, false /* strip non-digits */).toString();
   }
+  
+  /**
+   * @see PhoneNumberUtil#normalizeDigitsOnly(java.lang.CharSequence) 
+   */
+  public static String normalizeDigitsOnly(String number) {
+    return normalizeDigitsOnly((CharSequence) number);
+  }
 
   static StringBuilder normalizeDigits(CharSequence number, boolean keepNonDigits) {
     StringBuilder normalizedDigits = new StringBuilder(number.length());
@@ -749,6 +756,13 @@ public class PhoneNumberUtil {
   public static String normalizeDiallableCharsOnly(CharSequence number) {
     return normalizeHelper(number, DIALLABLE_CHAR_MAPPINGS, true /* remove non matches */);
   }
+  
+  /**
+   * @see PhoneNumberUtil#normalizeDiallableCharsOnly(java.lang.CharSequence) 
+   */
+  public static String normalizeDiallableCharsOnly(String number) {
+    return normalizeDiallableCharsOnly((CharSequence) number);
+  }
 
   /**
    * Converts all alpha characters in a number to their respective digits on a keypad, but retains
@@ -756,6 +770,14 @@ public class PhoneNumberUtil {
    */
   public static String convertAlphaCharactersInNumber(CharSequence number) {
     return normalizeHelper(number, ALPHA_PHONE_MAPPINGS, false);
+  }
+  
+  /**
+   * @see PhoneNumberUtil#convertAlphaCharactersInNumber(java.lang.CharSequence) 
+   */
+  @Deprecated
+  public static String convertAlphaCharactersInNumber(String number) {
+    return convertAlphaCharactersInNumber((CharSequence)number);
   }
 
   /**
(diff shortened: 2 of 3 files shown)


yinzara:
I don't believe my commit has caused the build to fail as the build log just contains the word "null".  I think something else is wrong.

yinzara:
I don't believe my commit has caused the build to fail as the build log just contains the word "null".  I think something else is wrong.

yinzara:
There was actually a typo in my prior code that caused the tests to fail.  After reviewing the test cases as well, I removed the @Deprecated annotation from the added methods as it causes libraries to think they're calling methods that are deprecated when they pass strings to the methods.

penmetsaa:
@yinzara,

Thank you for using and taking the time to improve libphonenumber!

We do not need seperate overloading APIs to accept the String literals as String is a char sequence. We can pass a string whenever we need a CharacterSequence. As there is no additional value for keeping these overloading APIs, we have chosen to change the signature of existing APIs itself and it will not break any client code.

I'll close this as it's not an issue with the library. Please consider joining
the [discussion group](https://groups.google.com/forum/#!forum/libphonenumber-discuss) if you'd like to discuss this further.

yinzara:
I've requested to join but do not have the ability to post.

The issue is not with the overloading of the methods.  The issue is with compatibility of the library with code compiled against older versions.

In my specific example, I have a production application that uses the libphonenumber library running 8.6.0 and uses the 'parse' method of PhoneNumberUtil.  I get a report that we are rejecting a specific phone number as invalid. I see that the change that would make that number valid is in libphonenumber 8.8.2 metadata.   I replace my libphonenumber-8.6.0.jar with libphonenumber-8.8.2.jar.  I restart my application and it crashes because when it attempts to call the original "parse" method of PhoneNumberUtil because the method does not exist (as it was compiled against the prior version which took in a String and the new version takes in a CharSequence).

I am in a regulated environment where recompiling my application will require approval from a regulatory board and cost my company time and resources (and means a delay of at least a month).  I do not need approval when changing public versions of libraries.  Because of this, I have no option but to keep the original 8.6.0 version of the library or recompile my application and go through regulatory approval again.

Because the method signatures changed (even though they changed to an interface that the concrete class implements), code compiled against the original methods will be unable to upgrade to a version of the library without recompiling their library against the newer APIs.

This discourages users of the library from upgrading from prior version of the library as it requires them to recompile their applications and not be backwardly compatible.

yinzara:
I've put together an example to demonstrate my point.
See:
https://github.com/yinzara/libphonenumber-exp

If you run "./compile.sh" that will compile Example.java using libphonenumber-8.6.0.jar

You can then "./run-old.sh 7025251231" and see it correctly validates the number using libphonenumber-8.6.0.jar.

You can then "./run-new.sh 7025251231" and see it throws a NoSuchMethodException because the 'parse' method that takes a String no longer exists.

This means anyone upgrading from libphonenumber-8.6.0 and lower to libphonenumber-8.7.0 and higher will need to recompile their applications first (and some may not even have the original source code).

My pull request will make the library compatible without requiring dependent libraries to recompile.

penmetsaa:
oh ok. Got you :)

Generally, it is good practice to compile and run against same library/jar version.
APIs can not be added back for resolving this kind of compatibility issue.

Possible workaround:
- For time being, maybe you can do these changes in your local copy and build the jar file.
- In next official deployment you can compile your library against new libphonenumber jar file so that
 you need not to maintain your own version.

How to build jar files:
- cd libphonenumber/java
- mvn package
- ant jar

yinzara:
While I understand your perspective, "good practice" is exactly that, a suggested practice.  

With a library as widely used as libphonenumber, the authors should strive to be backwards and forwards compatible with all changes so that people using the library have the greatest flexibility especially when most changes to the library are metadata about the validity and structure of phone numbers (which implies the library needs to be regularly updated to keep up to date with current metadata).  This a central tenant of the java programming language (i.e. version compatibility).

Many times the library that compiled against libphonenumber is one that the user has no control off (i.e. another third party library that uses libphonenumber in which you utilize that library).   The change made in 8.7.0 is a "breaking change" that causes all prior versions of the library to be incompatible with future versions.   My pull request adds less than 1k to the size of the library and allows for the library to be backward and forward compatible with prior and future versions.

There are already a number of issues filed in different issue trackers on the web in which this problem has occurred outside of my use case:
https://github.com/fbsamples/account-kit-samples-for-android/issues/16
https://github.com/robolectric/robolectric/issues/3320 

I would like a consensus by the community as to reject to accept this patch beyond one contributor.

yinzara:
I have added an additional example in the above Git repo:
https://github.com/yinzara/libphonenumber-exp

In this example "awesome-shared-library" is an hypothetical shared java library controlled by another organization (whether it be another company not providing you the source or an open source project you do not control).   It depends on libphonenumber-8.6.0.  It uses libphonenumber to validate numbers passed to it before providing it's "awesome" functionality to the user of the library.

"my-app" is a very simple example java library with only a test case in which it depends on libphonenumber-8.8.12 (because it needs the updated metadata because of a change in some phone number regions) and calls a hypothetical "doAwesome" method with an example phone number.

The test case fails because of a runtime MethodNotFoundException.  It it uncorrectable unless the person who provides the "awesome-shared-library" is willing (and still available) to update the library to the newer version and provide an updated compilation.

With this pull request, this issue could have been completely avoided and the library be fully forward and backward compatible.

PR #2180, opened 2018-04-05
Added in geodata for North America using Local Calling Guide.

bradsoto:


bradsoto:
I signed it!

bradsoto:
Tonnes of unicode issues. Closing.

PR #2184, opened 2018-04-06
Update formatForMobileDialling() to return intl format for Uzbek numbers

penmetsaa:
This is to update formatForMobileDialling() to return international format instead of national format for Uzbek phone numbers.

UZ mobile and fixed-line numbers have to be formatted in international format or
prefixed with special codes like 03, 04 (for fixed-line) and 05 (for mobile) for
dialling successfully from mobile devices. As we do not have complete information on
special codes and to be consistent with formatting across all phone types returning
number in international format here.


penmetsaa:
Hi Andy,
Could you please review this PR again as I have updated java doc in my new commit?

PR #2190, opened 2018-04-17
Update geocoding and carrier config files

keghani:
I updated these via `ant -f java/build.xml junit`
This is also run in the metadata update script during release, so I'm not sure why they didn't get updated last with the v8.9.4 PR in https://github.com/googlei18n/libphonenumber/pull/2187

See history of these files:
https://github.com/googlei18n/libphonenumber/commits/master/java/carrier/src/com/google/i18n/phonenumbers/carrier/data/config
https://github.com/googlei18n/libphonenumber/commits/master/java/geocoder/src/com/google/i18n/phonenumbers/geocoding/data/config

This came up when I auto-generated files for an upcoming change I'm making.


keghani:
Please don't merge (yet). I think this is due to my previously using a different version of Java.

keghani:
Yep, it was my Java version. Master's up-to-date (except the 2 .jar files, I'll fix those separately).

PR #2191, opened 2018-04-17
Add typecast for prefixMatcher in JS library

keghani:
cl/192841937

No build artifacts are affected with this change.

`tools/java/cpp-build/target/cpp-build-1.0-SNAPSHOT-jar-with-dependencies.jar` and `tools/java/java-build/target/java-build-1.0-SNAPSHOT-jar-with-dependencies.jar` are already outdated in `master`, but I haven't checked whether those changes contain build tools, and I'm not addressing those here.


keghani:
Closing for now. If it becomes needed I'll re-open.

PR #2214, opened 2018-06-01
Netherlands 'Machine to Machine' long numbers

vladimir-radchuk:
https://nl.wikipedia.org/wiki/Lijst_van_Nederlandse_netnummers

0970 t/m 0978
Elektronische communicatiediensten voor geautomatiseerde toepassingen (M2M- Machine to Machine), deze nummers zijn 12 cijfers lang

### DIFF
--- a/resources/PhoneNumberMetadata.xml
+++ b/resources/PhoneNumberMetadata.xml
@@ -20330,12 +20330,17 @@
           <leadingDigits>[89]0</leadingDigits>
           <format>$1 $2</format>
         </numberFormat>
+        <numberFormat pattern="(97[0-8])(\d{8})">
+          <leadingDigits>97[0-8]</leadingDigits>
+          <format>$1 $2</format>
+        </numberFormat>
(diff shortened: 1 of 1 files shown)


andyst:
M2M numbers cannot be added as `<mobile>`.
More design work is needed to support M2M.
This is discussed / tracked at: https://issuetracker.google.com/issues/74493346

We cannot merge this PR. Thanks for the contribution and for understanding.

ralphm:
@andyst Dutch telcos also provide SIMs with these numbers for data-only use in tablets and laptops. E.g. T-Mobile NL has [MB-verdeler](https://www.t-mobile.nl/klantenservice/abonnement-en-prepaid/voordeelbundels/mb-verdeler) with long 097 numbers. The FAQ entry on M2M numbers mentions the use case of _allowing people to sign up for services_ to fall outside of M2M, which I think does apply to the use of 097 numbers in tablets in The Netherlands.

PR #2224, opened 2018-06-29
Fix Chicago timezone definition.

nicpottier:
Extra | in the line.


nicpottier:
I signed it! (the CLA that is)

nicpottier:
Closing since it looks like this was addressed.

kkeshava:
Hi,
Sorry for late reply. Yes we had fixed this as part of https://issuetracker.google.com/issues/110978455.

PR #2234, opened 2018-08-13
Moving stl helper functions into inner namespace (gtl::).

lararennie:
Updating references in code. Removing an unused using declaration.


RamGopalBishnoi:
I didnot understand about it if in hindhi

RamGopalBishnoi:
Trancelate in hindi

PR #2249, opened 2018-08-31
Check whether non-test JS files are ASCII-safe during CI

mathiasbynens:


mathiasbynens:
It's working: https://travis-ci.org/googlei18n/libphonenumber/builds/423047558#L5791

mathiasbynens:
> Nice to see it caught things :)

Only in code comments, which is probably fine (i.e. at least no run-time behavior would change if the file is read using the wrong encoding) :) But yeah, easier to just apply the “no non-ASCII” rule consistently.

PR #2252, opened 2018-09-04
Update Unleashed prefix

Timvde:
As of September 2018, Mobile Vikings will also distribute new numbers
with prefix 456. Since Stievie is not a mobile brand, it can be ignored
in the listing.

Source: I work at Unleashed :)


Timvde:
I signed it!

resources/carrier/en/32.txt
```diff
@@ -15,11 +15,10 @@
...
 # a subsidiary of Belgacom.
 # Mobistar has re-branded as Orange on 9 May 2016.
-# Unleashed has 3 brands Mobile Vikings, JIM Mobile and Stievie. Prefix 456
-# phone numbers are mentioned in contact-us page of JIM mobile carrier. So it is
-# more likely that it is assigned to 'JIM Mobile'
+# Unleashed has 2 mobile brands: Mobile Vikings and JIM Mobile. Both of these
+# brands distribute numbers with prefix 456.
 
-32456|JIM Mobile
+32456|Unleashed
```
kkeshava: Unleashed is the company and we prefer the brand names instead. As mentioned in the comment above prefix 456 seems to be assigned to JIM Mobile only https://jimmobile.be/nl/contact/ so 456 should be JIM Mobile rather than Unleashed.

resources/carrier/en/32.txt
```diff
@@ -15,11 +15,10 @@
 # 1994 by Belgacom, following the launch of the first mobile network.Proximus is
 # a subsidiary of Belgacom.
 # Mobistar has re-branded as Orange on 9 May 2016.
-# Unleashed has 3 brands Mobile Vikings, JIM Mobile and Stievie. Prefix 456
-# phone numbers are mentioned in contact-us page of JIM mobile carrier. So it is
-# more likely that it is assigned to 'JIM Mobile'
+# Unleashed has 2 mobile brands: Mobile Vikings and JIM Mobile. Both of these
```
kkeshava: Unleashed website https://unleashed.be/ still says that it has 3 brands 'Mobile Vikings, JIM Mobile and Stievie.'

resources/carrier/en/32.txt
Timvde: As stated in the commit message, Unleashed will very soon (probably this month) start distributing 456-prefixed phone numbers for Mobile Vikings too. A similar mention to that on the contact page can be found on [this FAQ page](https://support.vikingco.com/hc/en-us/articles/360000313477-What-will-happen-after-activation-on-the-new-platform-):

> The short number for voicemail remains the same at 1933. If you want to call your voicemail from abroad, bear in mind that the number now is +32 **456** 19 19 33 instead of +32 486 19 33 33.

Since the same prefix will be used for new onboardings for both brands (and MV is the bigger one, by the way), it is wrong to call it JIM Mobile. It would be fine to list "JIM Mobile/Mobile Vikings" if you think "Unleashed" is inappropriate.

resources/carrier/en/32.txt
Timvde: Also as stated in the commit message, Stievie is not a mobile carrier (it's a tv app) and has nothing to do with phone numbers at all. Note that I changed the wording from "3 brands" to "2 mobile brands".

AbdulmajidAlnouri: This is perfectly reasonable.

resources/carrier/en/32.txt
AbdulmajidAlnouri: Let's do "JIM Mobile/Mobile Vikings" as you said since they are the real underlying telecoms who are responsible for the range. Our preference is to avoid parent companies whenever possible.

Could you provide some external link as proof of this change and add it to the commit message too?

AbdulmajidAlnouri:
@Timvde First I wanna say thank you for your contribution!
I've responded to the open threads and hopefully we can merge this change in asap!
I just requested some external proof and an update the commit message before we finish but other than that it looks perfect.

AbdulmajidAlnouri:
Feel free to rename the branch to whatever you see fit to better capture the change too

Timvde:
Thank you for your response!

The branch name is still appropriate imo. Unleashed is clearly mentioned in the TXT file itself. But if you have a suggestion for a better name, I'm open to change it. Do you consider it important for your git history?

I'm currently on holiday (keeping track of emails, but that's about it). After the weekend, I'll rebase the branch, fix the conflict and apply the change from "Unleashed" to "Mobile Vikings/JIM Mobile" (At second thought, I propose to change the order, since Mobile Vikings is the bigger brand and will probably soon overtake JIM in 32456 numbers).

Is [this FAQ page](https://support.vikingco.com/hc/en-us/articles/360000313477-What-will-happen-after-activation-on-the-new-platform-) that I mentioned sufficient as proof in the commit message? As far as I know, there is currently no other public resource available mentioning the 32456 prefix. I could double-check that if you want to.

Timvde:
@AbdulmajidAlnouri Is it okay like this?

kkeshava:
Hi,
Thanks for making the changes. You are correct, no need to change the branch name.
Sounds good to us to keep Mobile Vikings first as it might overtake JIM.

Thank you for contributing to libphonenumber! The changes look good to us :)

Timvde:
What comments should I still follow up on? For me, this PR is ready to merge. If it isn't, please tell me what's missing.

PR #2258, opened 2018-09-10
added back Zapp network numbers

necenzurat:
Zapp was bought by Telekom. 
https://en.wikipedia.org/wiki/Zapp_Mobile

And for example, 07887xxxxx is still a valid number.

### DIFF
--- a/resources/carrier/en/40.txt
+++ b/resources/carrier/en/40.txt
@@ -12,10 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Prefixes 78[08] were previously assigned to zapp.ro (TELEMOBIL S.A.) but they
-# are no longer operational and we are not aware to which operator it is
-# currently assigned.
-
 407000|Enigma-System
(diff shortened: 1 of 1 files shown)


necenzurat:
I signed it!

AbdulmajidAlnouri:
Awesome and thank you for your contribution.

Could you please provide some sourced proof of this ownership and number range reallocation?
From the wiki page you linked (https://en.wikipedia.org/wiki/Zapp_Mobile) I see "As of 24 March 2013 Zapp CDMA licence expired" but I suppose Telekom was granted that pool/range since you say that range is still active. Some official documentation like IR.21 forms would be good to capture this.

I'll be happy to dig for those docs if I find the time.

Context we should capture:
* Number range reallocated to Telekom.
* Acquisition of Zapp Mobile by Telekom.

penmetsaa:
Hi Majid,
Sorry, missed to update this PR.
As we found supporting IR.21 doc, merging this PR.

PR #2006, opened 2017-11-13
Update GB (+44) carrier data from Ofcom

giggsey:
Most recent Ofcom carrier list generated from https://github.com/giggsey/GBCarrierGenerator

PR #2283, opened 2018-11-13
Regenerate GB carrier data

giggsey:
The recent release (v8.9.10) updated the GB Carrier data, however, it was still out of date.

This merge updates to the latest data from Ofcom, generated on 2018-11-13 using https://github.com/giggsey/GBCarrierGenerator

Replaces #2006

### DIFF
--- a/resources/carrier/en/44.txt
+++ b/resources/carrier/en/44.txt
@@ -13,26 +13,23 @@
 # limitations under the License.
 
 # Carrier details updated from http://www.ofcom.org.uk/static/numbering/S7.xls
-# Generated on 2017-07-14 using https://github.com/giggsey/GBCarrierGenerator
+# Generated on 2018-11-13 using https://github.com/giggsey/GBCarrierGenerator
 
 # Carrier names have been updated to friendlier trading/brand names.
 # The mapping of company names can be found below:
 #  - (aq) Limited trading as aql: aql
 #  - 08Direct Limited: 08Direct
-#  - 09 Mobile Ltd: 09 Mobile
 #  - 24 Seven Communications Ltd: 24 Seven
-#  - AMSUK LTD: AMSUK
 #  - Ace Call Limited: Ace Call
 #  - Airwave Solutions Ltd: Airwave
-#  - Alliance Technologies LLC: Alliance
 #  - Andrews & Arnold Ltd: Andrews & Arnold
 #  - Anywhere Sim Limited: Anywhere Sim
 #  - BT OnePhone Limited: BT OnePhone
 #  - Bellingham Telecommunications Limited: Bellingham
+#  - Bluewave Communications Limited: Bluewave Communications
 #  - CFL Communications Limited: CFL
 #  - Cheers International Sales Limited: Cheers
 #  - Citrus Telecommunications Ltd: Citrus
-#  - Cloud9 Communications Limited: Cloud9
 #  - Cloud9 Mobile Communications Ltd: Cloud9
 #  - Compatel Ltd: Compatel
 #  - Confabulate Limited: Confabulate
@@ -53,13 +50,11 @@
 #  - Hutchison 3G UK Ltd: Three
 #  - IPV6 Limited: IPV6
 #  - IV Response Limited: IV Response
-#  - Icron Network: Icron
 #  - Icron Network Limited: Icron Network
 #  - JT (Guernsey) Limited: JT
 #  - JT (Jersey) Limited: JT
 #  - Jersey Airtel  Limited: Airtel
 #  - Lanonyx Telecom Limited: Lanonyx
-#  - LegendTel LLC: LegendTel
 #  - Limitless Mobile Ltd: Limitless
 #  - Lleida.net Serveis Telematics Limited: Lleida.net
 #  - Lycamobile UK Limited: Lycamobile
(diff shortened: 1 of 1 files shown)


penmetsaa:
Hi,

Thanks so much for the contribution.

You have reverted some of the changes we did to this file during release 8.10.0, PR #2276
Recently we had automated the process of compressing / being specific in carrier metadata.
This will not effect O/P of API but did to follow consistent approach for maintaining all
carrier and geocode mappings.
Eg: We are now being specific like "447426|Three" to make it clear.

Could you please revert back those rollforwards and only place potential changes?

hagbard:
Since the GB data is now generated from a new internal representation, the mapping files are no longer editable directly, they are purely automatically generated. As such we can no longer accept any pull requests for these files.

The good news is that we should be able to easily merge any new carrier data into our data and regenerate new, consistent/minimal versions of the mapping files. The only time a mapping would not appear in our generated files is if the prefix was not one we considered a valid range (but since your source looks to be good, that would indicate that we might have missing ranges, which could then be added).

The best thing to do is provide us with your original data for carrier mappings (e.g. Offcom data) in a machine readable way, which we can then process and merge into our base data to regenerate correct mapping files. We can handle merging mapping files in the "prefix|carrier name" format, so that shouldn't be an issue. It looks like the data you have fits this bill.

This also means we should no longer show these files as being directly generated from your project (since they aren't anymore) but we're happy to continue to credit you for the source of the data (i.e. we should make a small adjustment to the comment at the top of this file).

Also, just to explain, the mapping files we generate are always minimal now, so:

4477442|Core Communication
4477443|Core Communication
4477444|Core Communication
4477445|Core Communication
4477446|Core Communication
4477447|Core Communication
4477448|Core Communication
4477449|Core Communication

would never exist in our data, only:

447744|Core Communication

(or perhaps even a shorter prefix). This shortening can only be done with access to all the range data, so as not to "over shorten" ranges. This is fine, since it returns exactly the same result for all valid numbers, but means that you will now always have a diff between the "unshortened" mappings from someone like Offcom and the mappings we generate.

Funnily enough I was just about to send out an announcement email explaining some of this to potential contributors, since I didn't want anyone spending time working on exact patches we can no longer accept (we still gladly accept data sources, just not pull requests to modify data files directly).

giggsey:
@hagbard That's fair enough.

The data itself comes from a machine readable format supplied by Ofcom (spreadsheet). I just have some basic PHP that converts it into the format for the metadata files, and deals with renaming companies & compressing the data.

The data in the merge request is compressed as far as it can safely go based on the current Ofcom allocations.

In your example of Core Communications, the shortened 447744 shouldn't be used, as Core Communications have only been allocated 2-9 (not 0 or 1).

In @penmetsaa's example above, Three has been compressed, because they have been allocated all the other ranges apart from the others specified.

> 44742|Three
> 447420|Orange
> 447421|Orange
> 447422|Orange
> 447423|Vodafone
> 447424|Lycamobile
> 447425|Vodafone

giggsey:
> The 0/1 ranges just aren't valid so we don't care what we return for them.

They validate as valid numbers though: http://libphonenumber.appspot.com/phonenumberparser?number=%2B447744012345.

This would be fine, but would require the metadata to be much more accurate than it is now (actual allocated numbers instead of possibly allocated numbers).

According to [Ofcom](https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbering-data) they try to publish data every Wednesday. Carrier Updates certainly aren't that important to be updated so frequently, but if the metadata is being mixed with the compression mentioned above, then that's sure to cause problems.

Keeping with this same number range as an example 77440 is currently not allocated, and currently does not report a carrier. It does however validate as a valid number. If the carrier data was compressed to just '7744': Carrier, and another CP was allocated 77440, then incorrect carrier information would start being shown.

My gmail address is the same as my GitHub username.

hagbard:
> > The 0/1 ranges just aren't valid so we don't care what we return for them.
> 
> They validate as valid numbers though: http://libphonenumber.appspot.com/phonenumberparser?number=%2B447744012345.

Oh it's possible that the specific example I cited isn't true, I was just trying to illustrate that when validity is known we can do a better job of simplifying the files. The 7744[01] ranges are indeed valid, and if you look in the file we don't compress those prefixes.

> This would be fine, but would require the metadata to be much more accurate than it is now (actual allocated numbers instead of possibly allocated numbers).

There's no issue, honestly. This migration has happened for hundreds of regions without issues in the mapping files, and the generation of mapping files is (a) correct and (b) minimal.

> According to [Ofcom](https://www.ofcom.org.uk/phones-telecoms-and-internet/information-for-industry/numbering/numbering-data) they try to publish data every Wednesday. Carrier Updates certainly aren't that important to be updated so frequently, but if the metadata is being mixed with the compression mentioned above, then that's sure to cause problems.

It will not cause issues, and when we update ranges, we will now be more likely to keep carrier mappings in the same file.

> Keeping with this same number range as an example 77440 is currently not allocated, and currently does not report a carrier. It does however validate as a valid number. If the carrier data was compressed to just '7744': Carrier, and another CP was allocated 77440, then incorrect carrier information would start being shown.

Right, that's just because I picked an illustrative example before we got into the specific details. A real example is:

44773|O2
447733|Vodafone
447735|Three
447737|Three

Which is what's in the mapping file now (and was auto-generated).

> My gmail address is the same as my GitHub username.

I'll see if I can share a copy of the new data table so you can see it's not an issue.

hagbard:
You should be able to see the ranges in:
https://docs.google.com/a/google.com/spreadsheets/d/e/2PACX-1vS6BoJIIDM0GZ0E6cjq9P4yl-2W6SiiqTBd6-MvoiB8J8Mhhpb90WGpyHg_5tT7amGz1HJFtshZxS8B/pubhtml
with a Gmail account.

PR #2293, opened 2018-12-06
Update README.md

penmetsaa:
Demo version number update for release 8.10.2

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,7 +55,7 @@ Cream Sandwich).
 The [Java demo](https://libphonenumber.appspot.com/) is updated with a slight
 delay after the GitHub release.
 
-Last demo update: v8.10.1.
+Last demo update: v8.10.2.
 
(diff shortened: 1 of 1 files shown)


sutiwa:
+66952511115 Thailand

penmetsaa:
@sutiwa,

Thank you for using libphonenumber!

This is actually working as intended; +66952511115 is already marked by
our library as valid mobile number. To check the latest behavior for
yourself (we update metadata frequently), see the demo:
https://libphonenumber.appspot.com/phonenumberparser?number=%2B66952511115+

If there is a specific range that you think we should update and don't do so
in the demo highlighted above, please file a new issue with details. See our
contribution guidelines:
https://github.com/googlei18n/libphonenumber/blob/master/CONTRIBUTING.md

Then we may determine whether something is a real issue, and the details we need
to be able to fix it.

PR #2301, opened 2019-01-08
 Updating "commons-fileupload" version number from 1.3 to 1.3.2

penmetsaa:
Followup to PR #2298 as sub version number is not mentioned earlier. GitHub has suggested it to be >~1.3.2. So just making it to 1.3.2 and it will be in line with commons-io version.

### DIFF
--- a/java/demo/pom.xml
+++ b/java/demo/pom.xml
@@ -78,7 +78,7 @@
     <dependency>
       <groupId>commons-fileupload</groupId>
       <artifactId>commons-fileupload</artifactId>
-      <version>1.3</version>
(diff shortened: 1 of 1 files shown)


penmetsaa:
Thanks Andrew. The notifications are keep on changing. Firstly, they mentioned to have ~> 1.3.1 (same or greater than 1.3.1) and later 1.3.2. As per below page making it 1.3.3 is fine.
https://github.com/googlei18n/libphonenumber/network/alert/java/demo/pom.xml/commons-fileupload:commons-fileupload/open

Made the change and tested as per our process. Please review/approve same.

PR #2318, opened 2019-02-21
fix grammar

bennytheshap:
"Try to" is the correct construction (vs. "try and").


lararennie:
Change looks fine but will need the CLA signed.

bennytheshap:
No thanks. Entering into a contract for changes as minor as these doesn’t seem worth it to me. These changes are minor, so that it should be easy for you to make your own PR if you want.

penmetsaa:
Sure. I made same update in #2322.
Thanks so much for your contribution.

PR #2319, opened 2019-02-21
Add example for one country with several prefixes.

6temes:
The current exaple of Kosovo explains a situation created by a territorial dispute. 

In the case of Reunion Island, the system was designed to operate like this.


6temes:
Signed CLA!

PR #2323, opened 2019-02-25
Add falsehood about accessibility of phone numbers

mpomery:


mpomery:
I signed it!

kkeshava:
Hi,

Thank you for taking the time to improve libphonenumber!

Using phones for calls completely depends upon individual’s phone settings. These are generic issues which the caller should be aware of. As it is not related to / within the scope of our library, we would not be adding it to FALSEHOODS.md. Hence closing the PR.

mpomery:
Hey @kkeshava,

If it were re-worded to be more like "You can send a text message to any phone number", would it be a better fit?

Other examples include fax machines, mobile internet connections and IOT/Standalone devices like GPS trackers and eftpos terminals.

kkeshava:
Hi,
Sending a text message to a phone number also depends on each individual's choice. It's the same with the other examples. As mentioned earlier, this is not related to / within the scope of our library.

mpomery:
Sorry, I was referring to the already existing Falsehood titled "You can send a text message to any phone number".

https://github.com/mpomery/libphonenumber/blame/fae515072fa7a65ad9fe465ba4c8a15fc00c8219/FALSEHOODS.md#L77

If the falsehood was similar:

>1. **You can make a call to any phone number**
>  
>    Some devices such as EFTPOS terminals, fax machines and mobile internet dongles may not support receiving calls. In addition, some people can not use their phones for phone calls. This may be permanent (such as a hearing disability), temporary (temporary hearing loss) or situational (when the user is in a noisy environment).
>  
>    **Tip:** Make how you use phone numbers apparent to the user and give them the option to set their preference. Ensure that you look up this preference before attempting to communicate with the user.

kkeshava:
Hi,
Sorry had not understood correctly about your previous comment :)
Your last suggestion seems perfect could you please make the changes accordingly, I shall approve it.

mpomery:
Branch updated 😊

PR #2349, opened 2019-05-16
'animal-sniffer-maven-plugin' changes: version is upgraded and extracted into root pom.xml

dejan2609:
Note: this is a maven build improvement.

Details:
- plugin version is upgraded: **1.15** -->> **1.17**
- plugin version is removed out of submodules into root pom.xml (i.e. plugin version can be changed easily)

PR #2356, opened 2019-06-12
Reverting animal sniff plugin version update PR #2349 and release PR of 8.10.14 #2354 

penmetsaa:
As per my observations based on error stack and googling the things:
- Animal sniff plugin is a bit strict on getting the signatures right especially when the complied java version is different than the target version and there is method signature changes.
- Error: java.lang.NoSuchMethodError: java.nio.ByteBuffer.mark()Ljava/nio/ByteBuffer;
- Maybe these are existing issues but there were caught now at compile time after we upgraded the plugin.
- As this is blocking the Maven release for version 8.10.14, reverting the release commit as well as it contains new dependency jars generated. Eg: tools/java/java-build/target/java-build-1.0-SNAPSHOT-jar-with-dependencies.jar


lararennie:
Is the issue is that we didn't aim to have everything compiled at 1.5, i.e. the demo wasn't supposed to be 1.5 specific, and the upgrade applied it to the whole project?

penmetsaa:
I am not sure Lara. It is not very clear in the stacktrace which I shared to you now.
AFAIK all the modules are now compiled at java 8 and target is also at java 8. I guess some (as per google search results) dependency is compiled at higher version (maybe at 1.9) and we running at java 1.8 but it is unlikely to happen. Sorry, I am not very confident about my analysis as I do not have much prior experience in this area. So I thought of triaging this as a separate issue and unblock release first.

Any reason why you mentioned 1.5?

lararennie:
Makes sense.
Animal sniffer is there to make sure that the code is kept at Java1.5 SDK
(you can compile it at whatever though) for compatibility with old Android
versions - so we don't accidentally start using methods or classes added
later. However, the demo we didn't care about because it's on appengine. My
theory was that it might be failing because of that, but I'll look at the
stacktrace now.

ons. 12. jun. 2019 kl. 17:38 skrev penmetsaa <notifications@github.com>:

penmetsaa:
Thanks Lara.
Just FYI. We were not able to proceed with release even after rolled back i.e we are facing same issue. We can discuss further in b/135157237

PR #2359, opened 2019-06-18
Libphonenumber java jars are now compiled at java v1.6 (previously v1.5)

penmetsaa:
- Due to lack of support for older Java versions in Maven, we are now going to build our Java libphonenumber jars as Java 1.6, instead of Java 1.5.
- We will keep our API usage to 1.5 at this time for anyone compiling on their own


penmetsaa:
> Used pkgdiff, jars are identical, so seems changes are limited to version only, which is good :)

Thanks Lara :)

PR #2363, opened 2019-07-09
Add new build options

Venemo:
This pull request adds two new build options:

* `USE_PROTOBUF_LITE` - When this is set to `ON` then it links against `protobuf-lite` instead of the full version of `protobuf`. As far as I could see the metadata has `option optimize_for = LITE_RUNTIME;` so it should be safe to use the lite version. This is useful when you want to save some disk space. The default is `OFF`.
* `REGENERATE_METADATA` - When this is set to `OFF` it will skip regenerating the metadata with `BuildMetadataCppFromXml`. Since the metadata is included in the source tree anyway, it is beneficial for packagers to turn this `OFF`: it saves some time, and it also makes it unnecessary to have java in the build environment. The default is `ON`.

Please let me know what you guys think.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -86,6 +86,7 @@ option ("USE_LITE_METADATA" "Use lite metadata" "OFF")
 option ("USE_RE2" "Use RE2" "OFF")
 option ("USE_STD_MAP" "Force the use of std::map" "OFF")
 option ("BUILD_STATIC_LIB" "Build static libraries" "ON")
+option ("USE_PROTOBUF_LITE" "Link to protobuf-lite" "OFF")
 
 if (${USE_ALTERNATE_FORMATS} STREQUAL "ON")
   add_definitions ("-DI18N_PHONENUMBERS_USE_ALTERNATE_FORMATS")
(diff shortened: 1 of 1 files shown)


penmetsaa:
Can you break this up into two separate PRs, one for each option?

Re USE_PROTOBUF_LITE: we need to evaluate this change thoroughly before making any changes. Can you prepare a short one-pager doc discussing (a) the name of the option (taking into account that USE_LITE_METADATA already exists and means something different) (b) metrics on how much the difference in size is (c) what the difference in API is if USE_PROTOBUF_LITE is used?

Venemo:
@penmetsaa Sure think, I will split this into two separate PRs. About PROTOBUF_LITE, I'm not a protobuf expert, but sure I can try to summarize why.

gvollant:
> @penmetsaa Sure thing, I will split this into two separate PRs. About PROTOBUF_LITE, I'm not a protobuf expert, but sure I can try to summarize why.

I did not find your two PR @Venemo

gvollant:
@Venemo @penmetsaa I think the REGENERATE_METADATA can be very useful (I needed install java on some system only to compile libphonenumber), and we must integrate it.

gvollant:
@Venemo @penmetsaa 
I created #2475 with only the REGENERATE_METADATA option

Venemo:
@gvollant  I'm afraid I haven't got around to doing this yet, thanks for doing it

penmetsaa:
Hi Venemo and gvollant,
Thanks so much for your work on these build options.
I have approved the new PR regarding skipping of generation of metadata.

@Protobuf-lite,
Please reopen this PR or raise a case at: https://issuetracker.google.com/issues?q=componentid:192347%20status:open
if you can articulate the advantages of it over existing build conf "USE_LITE_METADATA".

gvollant:
regenerate metadata merged with #2475

protobuf on separate pr #2482 

@Venemo @penmetsaa

gvollant:
@Venemo can you take a look at #2482 discussion ?

gvollant:
There is a very good documentation at #2482

gvollant:
#2475 and #2482 are now merged

PR #2367, opened 2019-07-26
Add ability to compile metadata from string of bytes

harshpa1:
**Changes Made:**
- Added ability to write binary metadata into .pb files by passing CREATE_PB_FILES=ON into CMake.
- Added ability to create a phonenumberutil object through passing in a function pointer that creates a string which corresponds to the bytes of compiled in metadata.

**Reasons for Change:**

C++ metadata files are very large in size (~ 1 MB) and thus users of libphonenumber can benefit from being able to compress the compiled in metadata. By adding the ability to both generate binary files and pass in a string corresponding to the files, libphonenumber users can use their own compression tools to save binary size.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -84,6 +84,7 @@ option ("USE_BOOST" "Use Boost" "ON")
 option ("USE_ICU_REGEXP" "Use ICU regexp engine" "ON")
 option ("USE_LITE_METADATA" "Use lite metadata" "OFF")
 option ("USE_RE2" "Use RE2" "OFF")
+option ("CREATE_PB_FILES" "Create .pb binary metadata files" "OFF")
 option ("USE_STD_MAP" "Force the use of std::map" "OFF")
 option ("BUILD_STATIC_LIB" "Build static libraries" "ON")
 
@@ -268,14 +269,17 @@ function (add_metadata_gen_target TARGET_NAME
   set (METADATA_SOURCE_DIR "${CMAKE_SOURCE_DIR}/src/phonenumbers")
   set (GEN_OUTPUT "${METADATA_SOURCE_DIR}/${METADATA_TYPE}.cc"
                   "${METADATA_SOURCE_DIR}/${METADATA_HEADER}.h")
+  if(CREATE_PB_FILES STREQUAL "ON")
+    list(APPEND GEN_OUTPUT "${METADATA_SOURCE_DIR}/${METADATA_TYPE}.pb")
+    set (CREATE_PB_FILES "TRUE")
+  endif()
   set (JAR_PATH "${CMAKE_SOURCE_DIR}/../tools/java/cpp-build/target")
   set (JAR_PATH "${JAR_PATH}/cpp-build-1.0-SNAPSHOT-jar-with-dependencies.jar")
 
   add_custom_command (
     COMMAND ${JAVA_BIN} -jar
       ${JAR_PATH} BuildMetadataCppFromXml ${XML_FILE}
-      ${CMAKE_SOURCE_DIR}/src/phonenumbers ${METADATA_TYPE}
-
+      ${CMAKE_SOURCE_DIR}/src/phonenumbers ${METADATA_TYPE} ${CREATE_PB_FILES}
     OUTPUT ${GEN_OUTPUT}
     DEPENDS ${XML_FILE}
   )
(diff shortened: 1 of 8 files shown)


tools/java/cpp-build/src/com/google/i18n/phonenumbers/BuildMetadataCppFromXml.java
```diff
@@ -215,6 +248,10 @@ OutputStream openSourceStream(File dir, Type type, Variant variant) throws FileN
     return new FileOutputStream(new File(dir, variant.getBasename(type) + ".cc"));
   }
 
+  OutputStream openOutputStream(File dir, Type type, Variant variant) throws FileNotFoundException {
```
SamMaier: Perhaps a slightly better name would be "openRawProtoStream"

cpp/CMakeLists.txt
```diff
@@ -268,14 +269,17 @@ function (add_metadata_gen_target TARGET_NAME
   set (METADATA_SOURCE_DIR "${CMAKE_SOURCE_DIR}/src/phonenumbers")
   set (GEN_OUTPUT "${METADATA_SOURCE_DIR}/${METADATA_TYPE}.cc"
                   "${METADATA_SOURCE_DIR}/${METADATA_HEADER}.h")
+  if(CREATE_PB_FILES STREQUAL "ON")
+    list(APPEND GEN_OUTPUT "${METADATA_SOURCE_DIR}/${METADATA_TYPE}.pb")
+    set (CREATE_PB_FILES "TRUE")
```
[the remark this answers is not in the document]
harshpa1: Done.

cpp/src/phonenumbers/phonenumberutil.cc
```diff
@@ -753,20 +762,23 @@ PhoneNumberUtil::PhoneNumberUtil()
...
+// Initializes PhoneNumberUtil fields by using data from PhoneMetadataCollection.
+void PhoneNumberUtil::InitializeFromMetadata(PhoneMetadataCollection& metadata_collection){
   // Storing data in a temporary map to make it easier to find other regions
   // that share a country calling code when inserting data.
   std::map<int, std::list<string>* > country_calling_code_to_region_map;
   for (RepeatedPtrField<PhoneMetadata>::const_iterator it =
-           metadata_collection.metadata().begin();
-       it != metadata_collection.metadata().end();
-       ++it) {
+          metadata_collection.metadata().begin();
```
SamMaier: nit: spaces here don't align with previous code - I'm not totally sure which one is correct.

cpp/src/phonenumbers/phonenumberutil.cc
```diff
@@ -806,12 +818,9 @@ PhoneNumberUtil::PhoneNumberUtil()
   // Sort all the pairs in ascending order according to country calling code.
   std::sort(country_calling_code_to_region_code_map_->begin(),
             country_calling_code_to_region_code_map_->end(), OrderByFirst());
-}
 
-PhoneNumberUtil::~PhoneNumberUtil() {
-  gtl::STLDeleteContainerPairSecondPointers(
-      country_calling_code_to_region_code_map_->begin(),
-      country_calling_code_to_region_code_map_->end());
+  // Sets has_used_metadata_ to true for GetInstance()
```
SamMaier: This comment would be more useful if it mentioned that this prevents the metadata from being read two times in GetInstance.

cpp/src/phonenumbers/phonenumberutil.cc
```diff
@@ -874,7 +883,39 @@ void PhoneNumberUtil::GetSupportedTypesForNonGeoEntity(
...
+    }
+    phone_number_util->InitializeFromMetadata(metadata_collection);
+  }
+
+  return phone_number_util;
+}
+
+// Public wrapper function to get a PhoneNumberUtil instance with a string of
+// bytes. Takes as input a function that returns the processed metadata as a 
+// string. Most common use case for this function is to decompress metadata.
```
SamMaier: Most common -> one possible

SamMaier:
I am going to take this over so if we need to make changes we can, since Harsh's internship is done. Thus, this pull request can be abandoned. Since Harsh's internship is done, so someone with permissions might have to abandon it for him.

SamMaier:
This has been split into 2 parts:

Adding support for metadata proto output to .pb file
https://github.com/google/libphonenumber/pull/2384

Adding support to load metadata via function wrapper
https://github.com/google/libphonenumber/pull/2385

penmetsaa:
Closing this PR as it has been moved to a different PRs.

PR #2371, opened 2019-08-01
Lock C++11 implementation based on std::mutex.

dvadym:
libphonenumber doesn't have Lock implementation on Windows. This pull request implements Lock with C++11 std::mutex. As a a result libphonenumber would be thread-safe when build under C++11.


dvadym:
Another option is to remove platform specific Lock implementation completely, leave only based on std::mutex. It looks that it wouldn't narrow platforms where it can be built, since std::unique_ptr (which is also C++ feature) is already used in libphonelibrary https://github.com/google/libphonenumber/blob/master/cpp/src/phonenumbers/phonenumbermatcher.cc#L704

To reviewers: please advice, what do you think is better?

gvollant:
Hello,
this is a good idea, but there is no support of Singleton class

I added it on #2400 , and also code with Win32 native api with CriticalSection

https://github.com/google/libphonenumber/pull/2400/commits/228a04caefeea7bc9758b5311a34955acf0d93fe

PR #2376, opened 2019-08-08
Abolish MX's international mobile token (1)

penmetsaa:
Remove MX's mobile token value (1) from mobileTokenMap list in PhoneNumberUtil.
Update AYTF to not swallow input numbers like mobile token (1), which we are trimming of when asked to format, as AYTF should represent exact input digits. (This API is only in Java and CPP)

MX's mobile token value (1) has been abolished as part of dialling pattern updates in MX, during Aug 2019. https://www.itu.int/dms_pub/itu-t/oth/02/02/T020200008A0003PDFE.pdf
More details in cl/261295217 and cl/263086682

Notes: 
- Still using HashMap, not ImmutableMap.of() as it needs java 8 and to be consistent with other part of the existing code.
- This PR will be merged on after success completion of migration.
- Will update formatForMobileDialling() for MX case (api that returns most successful format of dialling phonenumbers from mobile), where presently it returns E.64 format for MX and for other regions national format. After this migration national format should hold good enough (previously it failed in some cases based on test results) for mobile dialling within the region.

### DIFF
--- a/cpp/src/phonenumbers/asyoutypeformatter.cc
+++ b/cpp/src/phonenumbers/asyoutypeformatter.cc
@@ -479,8 +479,24 @@ void AsYouTypeFormatter::AttemptToFormatAccruedDigits(
       DCHECK(status);
       IGNORE_UNUSED(status);
 
-      AppendNationalNumber(formatted_number, formatted_result);
-      return;
+      string full_output(*formatted_result);
+      // Check that we didn't remove nor add any extra digits when we matched
+      // this formatting pattern. This usually happens after we entered the last
+      // digit during AYTF. Eg: In case of MX, we swallow mobile token (1) when
+      // formatted but AYTF should retain all the number entered and not change
+      // in order to match a format (of same leading digits and length) display
+      // in that way.
+      AppendNationalNumber(formatted_number, &full_output);
+      phone_util_.NormalizeDiallableCharsOnly(&full_output);
+      string accrued_input_without_formatting_stdstring;
+      accrued_input_without_formatting_.toUTF8String(
+          accrued_input_without_formatting_stdstring);
+      if (full_output == accrued_input_without_formatting_stdstring) {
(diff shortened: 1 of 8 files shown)


penmetsaa:
@Avelino187,
Are you saying that the number you have shared is already reachable without any additional international MX mobile token 1? We will be releasing metadata update for MX in upcoming release, mostly by end of this week. Please let me know if there are any issues (like validity checks) other than formatting in our demo portal. More details regarding any update of this MX metadata rollout will be published at http://issuetracker.google.com/issues/137580716 and the discussion
group (https://groups.google.com/forum/#!forum/libphonenumber-discuss)

hagbard:
I don't understand the comment about Java8? ImmutableMap most certainly does not require Java 8 to be present. Can you point me at the code you're talking about?

avimar:
Is there a way to automagically remove the previous required 1?

lararennie:
> I don't understand the comment about Java8? ImmutableMap most certainly does not require Java 8 to be present. Can you point me at the code you're talking about?

I think he means guava.

lararennie:
> Is there a way to automagically remove the previous required 1?

You can format E164 -> parse (once metadata has been updated)
New number is now correct.

penmetsaa:
@hagbard,
The code is in cl/263086682. The open-source version is at 1.6 and as per google search the initialisation of ImmutableMap through of() is available only after guava style inclusion in Java 8.
And also in open source version we arre using HashMap so continued same. I may got it wrong, please suggest, will edit the description accordingly.

penmetsaa:
@avimar 
We made [announcement](https://groups.google.com/forum/#!topic/libphonenumber-discuss/pecTIo_HpVE) about this metadata change at libphonenumber-discuss@googlegroups.com. If you are not part of it already, please join this group.

avimar:
thanks @penmetsaa, subscribed.

(Is there a way to subscribe only to those type of metadata updates and not all releases?)

PR #2395, opened 2019-10-01
Update protobuffer release version to 3.6.1

penmetsaa:
This is to stay relevant with new proto buffer releases and also rectify the compilations issues during the release - new proto buffer uses ByteBuffer instead of ByteString - b/141842640


lararennie:
Should you not also update the c++ readme to recommend a higher proto version too?

lararennie:
Also someone should jardiff the jars to check they only have the expected changes.

penmetsaa:
Hi Lara,
Checked jardiff and all diffs are due to protobuff. Not saved the diff of other jar.
https://docs.google.com/document/d/1Vi-aZl2-IswMp-vwiHnieDaenRnKRFCcyZPCni1Ox9k/edit

@ Changing  c++ readme,
I do not think it is needed because this dependency change is only for tools/java/cpp-build and I see no reference of old 3.0.0 version in CPP readme and talking about cpp-build. It mostly mentions something like "consider 2.4 or higher" and "some manual instructions if same GNU/Linux version is not available". Eg: tar xjf protobuf-2.4.tar.bz2 or tar xjf protobuf-3.6.1.tar.bz2.

But it throws errors when I run those commands "tar xjf protobuf-3.6.1.tar.bz2
tar (child): protobuf-3.6.1.tar.bz2: Cannot open: No such file or directory" for both versions. Please let me know where and how I can correct CPP README if the change is really needed. This I can send as a separate PR to unblock the release.

lararennie:
> Hi Lara,
> Checked jardiff and all diffs are due to protobuff. Not saved the diff of other jar.
> https://docs.google.com/document/d/1Vi-aZl2-IswMp-vwiHnieDaenRnKRFCcyZPCni1Ox9k/edit
> 
> @ Changing c++ readme,
> I do not think it is needed because this dependency change is only for tools/java/cpp-build and I see no reference of old 3.0.0 version in CPP readme and talking about cpp-build. It mostly mentions something like "consider 2.4 or higher" and "some manual instructions if same GNU/Linux version is not available". Eg: tar xjf protobuf-2.4.tar.bz2 or tar xjf protobuf-3.6.1.tar.bz2.
> 
> But it throws errors when I run those commands "tar xjf protobuf-3.6.1.tar.bz2
> tar (child): protobuf-3.6.1.tar.bz2: Cannot open: No such file or directory" for both versions. Please let me know where and how I can correct CPP README if the change is really needed. This I can send as a separate PR to unblock the release.

So yes it fails because it assumes you have downloaded the files first, if you haven't the commands make no sense.

The point being: we now run on 3.6 locally, so I doubt that it would work if someone truly tried to run on 2.4. I would up the required version to 3.6 to match the maven file.

penmetsaa:
Make sense. I am just worried that if I can place right installation instructions for this newer version :) Tried at #2399. Please review same.

PR #2401, opened 2019-10-06
spellcheck(small grammatical mistake resolved)

rajbheda5:
Changed 'dialed' to 'dialled' as grammatical rule states that we ‘double a single consonant letter at the end of any base where the preceding vowel is spelled with a single letter and stressed’.


rajbheda5:
@googlebot I've signed it.

PR #2411, opened 2019-11-13
Metadata updates for release 8.10.23

rohininidhi:
Nov 13, 2019: v8.10.23
Metadata changes:
 - Updated phone metadata for region code(s): AL, AR, XK
 - Updated carrier data for country calling code(s): 383 (en)


fpalamour:
Metadata for Gabon (GA, +241) need to be updated. They added a digit, increasing the length from 8 to 9 digits.

PR #2417, opened 2019-11-30
Ensure that the local headers have priority over installed headers

dfaure-kdab:
This fixes the build error I saw today:
phonenumbermatcher.cc:715:7: error: ‘GetNationalNumberGroupsForPattern’ was not declared in this scope
due to the compiler picking an older installed
phonenumbers/phonenumbermatcher.h header than the one under src.

This problem doesn't usually happen with cmake because normally
this would all be done from src/CMakeLists.txt with
set(CMAKE_INCLUDE_CURRENT_DIR ON), but that would require more
refactoring.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -347,7 +347,7 @@ if (NOT WIN32)
   add_definitions ("-Wall -Werror")
 endif ()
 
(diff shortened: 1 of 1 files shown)


dfaure-kdab:
@googlebot I signed it!

PR #2424, opened 2019-12-27
Added area code 672 (British Columbia)

Thundefined:
Added 672 area code for British Columbia phone numbers.
My number is +1 672 999 **** and I can not add this number to my google account

This area code was added on May 4, here is the proof: https://www.eastlink.ca/telephone/newareacode

### DIFF
--- a/resources/geocoding/en/1.txt
+++ b/resources/geocoding/en/1.txt
@@ -18827,6 +18827,7 @@
 1667|Maryland
 1669|California
 1671646|Tamuning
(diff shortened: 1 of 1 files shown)


Thundefined:
@googlebot I signed it!

kkeshava:
Hi,
This is a duplicate of https://issuetracker.google.com/issues/146166984. We have already fixed it but we are now in a release freeze period and won't be releasing this before 2nd week of Jan 2020.

PR #2445, opened 2020-03-11
Add JavaScript demo for AsYouTypeFormatter

kimsey0:
The repository contains a JavaScript version of the `AsYouTypeFormatter`, but doesn't show off actual as-you-type formatting in the demo. This pull request currently contains a naive implementation of as-you-type formatting of the phone number field in the demo when a new checkbox is checked.

One might actually want to, for example:
1. Move this out into a separate demo page.
2. Change the as-you-type input box to a design that indicates and enforces that it's _append only_, since the `AsYouTypeFormatter `doesn't support deleting digits that have already been input, changing digits, or adding digits in the middle of the phone number.
3. Ensure that JavaScript works with all browsers that the demo is supposed to support (which are those?) and change usage and placement of state variables `savedValue` and `formatter`.
4. Update the compiled version of the demo, or create a compiled version of the separate demo page.

However, I hope this pull request can at least shed light on whether an as-you-type formatting demo is desirable. The issue tracker at https://issuetracker.google.com/issues?q=componentid:192347 seems to only contain issues regarding metadata. Having read the contribution guidelines, I'm still unsure if I should file an issue there with regards to demo improvements like this.

### DIFF
--- a/javascript/i18n/phonenumbers/demo.html
+++ b/javascript/i18n/phonenumbers/demo.html
@@ -42,7 +42,11 @@ limitations under the License.
 <form>
   <p>
   Specify a Phone Number:
-  <input type="text" name="phoneNumber" id="phoneNumber" size="25" />
+  <input type="text" name="phoneNumber" id="phoneNumber" size="25" oninput="phoneNumberFormatter" />
+  </p>
+  <p>
+  <input type="checkbox" name="formatAsYouType" id="formatAsYouType" />
+  Format as you type
   </p>
   <p>
   Specify a Default Country:
--- a/javascript/i18n/phonenumbers/demo.js
+++ b/javascript/i18n/phonenumbers/demo.js
@@ -171,3 +171,50 @@ function phoneNumberParser() {
 }
 
 goog.exportSymbol('phoneNumberParser', phoneNumberParser);
+
+let savedValue = "";
+let formatter = null;
+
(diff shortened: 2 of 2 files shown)


kimsey0:
@googlebot I fixed it.

lararennie:
Re 2) Yes the AYTF does support this - but if we just want to put the numbers that were input, then indeed it would be append only. I haven't played with this JS demo, but IMO you should try and match the Java demo (libphonenumber.appspot.com).

Note that it takes the input from the phone number box and enters it one char at a time, including formatting: so if the user puts any spaces in for instance, AYTF stops. (We might want to add a sentence warning the user about this actually)

kimsey0:
@lararennie: Thanks for the quick response.

[The existing JavaScript demo](https://raw.githack.com/google/libphonenumber/master/javascript/i18n/phonenumbers/demo-compiled.html) already matches the Java one at libphonenumber.appspot.com. However, neither one does as-you-type formatting, they just show a history of how the number would have been formatted after inputting each character when the Submit button is clicked.

This pull request actually formats the user input in the input field as the user types it, which I think has more value as a demo.

(I'd be happy to create a separate pull request that warns about the space issue, but it would also be great to be able to disable that behavior of the AsYouTypeFormatter.)

lararennie:
Then can we make it a separate demo?
And see my comment about (2).

PR #2443, opened 2020-03-04
Updates the type annotations in phonenumberutil.js

Dig-Doug:
Updates the type annotations to indicate the inputs are required to be non-null.

Context: http://cl/296235173

PR #2447, opened 2020-03-18
Make JS util test changes to reflect the annotation updates.

penmetsaa:
Like not null parameters. This is followup to recent cl/300367316. (#2443)
More details in cl/301500293


javascript/i18n/phonenumbers/phonenumberutil_test.js
```diff
@@ -3903,3 +3903,4 @@ function testIsAlphaNumber() {
   assertFalse(phoneUtil.isAlphaNumber('1800 123-1234 extension: 1234'));
   assertFalse(phoneUtil.isAlphaNumber('+800 1234-1234'));
 }
+
```
lararennie: nit: don't need extra blank line here.

PR #2466, opened 2020-05-15
Refer using std::string instead of global 'string'

penmetsaa:
Refer using std::string instead of global 'string' i.e to increase portablility, readability and also consistent. Cloning from cl/311566086

### DIFF
--- a/cpp/src/phonenumbers/utf/unicodetext.cc
+++ b/cpp/src/phonenumbers/utf/unicodetext.cc
@@ -27,6 +27,7 @@
 namespace i18n {
 namespace phonenumbers {
 
+using std::string;
 using std::stringstream;
 using std::max;
 using std::hex;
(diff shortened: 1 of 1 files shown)


cpp/src/phonenumbers/utf/unicodetext.cc
```diff
@@ -505,11 +505,11 @@ UnicodeText::const_iterator UnicodeText::MakeIterator(const char* p) const {
   return const_iterator(p);
 }
 
-string UnicodeText::const_iterator::DebugString() const {
+std::string UnicodeText::const_iterator::DebugString() const {
```
lararennie, line 508: Can we just use a using statement? It is the .cc file.

penmetsaa: Yah true that even better option. Done and tested.

PR #2471, opened 2020-05-20
[Japan] Fix wrong prefecture name in geocoding in Japan

eternalharvest:
## Abstract
I found a wrong prefecture name in geocoding resource file in Japan (+81).
Sasebo is a city in Nagasaki pref., and I think the prefix number is not special.
So, I think we need to change "Japan" => "Nagasaki".

I'm not familiar with this codebase, so I'm not sure if I can edit this resource file directly.

## Metadata
*   Country: Japan
*   Example number(s) and/or range(s): +81956
*   Number type ("fixed-line", "mobile", "short code", etc.): fixed-line
*   For short codes, cost and dialing restrictions: n/a
*   Where or whom did you get the number(s) from: n/a
*   Authoritative evidence (e.g. national numbering plan, operator announcement): https://en.wikipedia.org/wiki/List_of_dialing_codes_in_Japan  
https://www.soumu.go.jp/main_content/000141817.pdf (Japanese)  
According to the goverment document above, the prefix number is mainly used in Sasebo city in Nagasaki pref.
*   Link from demo (http://libphonenumber.appspot.com) showing error: n/a

### DIFF
--- a/resources/geocoding/en/81.txt
+++ b/resources/geocoding/en/81.txt
@@ -1126,7 +1126,7 @@
 819556|Karatsu, Saga
 819557|Karatsu, Saga
 819558|Karatsu, Saga
-81956|Sasebo, Japan
+81956|Sasebo, Nagasaki
 819572|Isahaya, Nagasaki
 819573|Isahaya, Nagasaki
 819574|Isahaya, Nagasaki


eternalharvest:
I signed the CLA.

kkeshava:
Hi,

We are not accepting pull requests for such metadata updates at this time. See our contribution guidelines: https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md.

We have created a bug in our Issue tracker: https://issuetracker.google.com/issues/157111072 Please check it for further updates.

eternalharvest:
@kkeshava 

Thank u for creating issue!

By the way, I wondering if I can use custom resources on my own projects.
Is there a way to switch java resources with this library in locale aware way?
Or need to write some patch to achieve this?

Best regards.

penmetsaa:
Hi,

Locale aware way? maybe not, but you can do on a whole.
You can modify any metadata resources and generate the final binaries.
After necessary edits, you can do something like this:

```shell
$ cd tools/java/ && mvn package &> "$ERROR_FILE"
$ ant -f java/build.xml build-phone-metadata build-test-metadata \
      build-short-metadata build-alternate-metadata build-timezones-data \
      build-timezones-test-data
```

Note: It is recommended to raise an issue with us so that
the fix applied to master source than branching away with custom edits.
It also adds maintainance when you are merging new changes.

eternalharvest:
@penmetsaa
Thank you for tell me how to build metadata.

I'm not intended to fork this project.
Just I want to get prefecture name in Japanese locale too.

But only English locale has the prefecture name in Japan region, and it seems each records is not corresponding to Japanese one. And I'm not sure why these difference is. I'm unconfident if i should make a patch to overwrite Japanese locale so that we can get prefecture name.

Please tell me if u know about any reason why these metadata has difference, and whether if it is intended that prefecture name is not in Japanese locale. I want to write a patch for this project, but I want to know if there is historical/techinical reason for that or some decision made in the past.

Anyway, I appreciate any achievement and contribution in this project.
Thank you.

penmetsaa:
If you see any issue in our [demo website](https://libphonenumber.appspot.com/), please follow our contribution guidelines:
https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md

PR #2489, opened 2020-06-22
Update CountryCodeToRegionCodeMap.java

xcg-star:
add missing region code


penmetsaa:
Hi,

Thanks so much for taking time to improve the library.

Actually we do not support some regions ascribed to lack of phone number regulations in these. 
It is documented here: https://github.com/google/libphonenumber/blob/master/FAQ.md#why-are-bouvet-island-bv-pitcairn-island-pn-antarctica-aq-etc-not-supported.

The library's intension is to not support all regions but only help with phone number handling wherever it applies. If you have official data that we do not have, please check our contribution guidelines. https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md#filing-a-metadata-issue

PR #2491, opened 2020-07-05
Checkin servlet-api-2.5 jar and update demo ant scripts to compile java classes at 1.8

penmetsaa:
- Demo portal is hosted in Google appengine. As the new appengine-java-sdk do not include any web-serving framework, we need to check-in the servlet-api-2.5 jar seperately in java/demo/war/WEB-INF/lib, similar to commons packages. More details in b/160522611 and cl/319663890.

- As demo portal is not using the latest java features, having a cap on it, similar to what we have in Maven.


DimStar77 (commented):
servlet-api-2.5 jar is packed with non-free / non-distributable license boiler plates

e.g.

servlet-api-2.5/javax/servlet/resources/web-app_2_2.dtd

```
This product or document is protected by copyright and distributed
under licenses restricting its use, copying, distribution, and
decompilation.  No part of this product or documentation may be
reproduced in any form by any means without prior written authorization
of Sun and its licensors, if any.  
```

PR #2513, opened 2020-08-26
Use newer Closure deps builder in JS build

kjin:
`calcdeps.py`, which is being used in the JS build, has been deprecated since ~2012. Recently, a user reported in https://github.com/google/closure-library/issues/1084 that this caused problems when building the demo, because it doesn't handle `goog.module` (to which Closure Library has been migrating recently).

This PR fixes this by replacing `calcdeps.py` with `closurebuilder.py`, which supercedes it. `closurebuilder.py` itself is quite old, and is not the preferred way to build, so this is only a stopgap solution that may break again in the future.

In terms of other behavioral differences between the two Python scripts, the flags are extremely similar, except that some are renamed. The newer `closurebuilder.py` also requires that files have either a `goog.provide` or `goog.module` call, so those have been added to the various test files that don't have them.

Fixes https://github.com/google/closure-library/issues/1084


kjin:
Hi @penmetsaa, any progress on this PR?

penmetsaa:
Thanks Kelvin. Everything worked out well.
For posterity:
- After this change, JS build needs upto date closure-compiler and library.
- closure-compiler needs Blazel to build its jar so that libphonenumber can compiler against it.
More details mentioned in b/166863265

penmetsaa (approved):

PR #2520, opened 2020-09-13
Disallow N11 in US central office codes

dracos:
I realise https://issuetracker.google.com/issues/75404127 was closed, but I believe possibly incorrectly. I think N11 has to be disallowed in central office codes as well as area codes, because otherwise how would you dial them locally in places where an area code is still optional.

As per the commit message, if you download one of the Excel spreadsheets for CO Codes from https://www.nationalnanpa.com/reports/reports_cocodes_assign.html e.g. https://www.nationalnanpa.com/nanp1/CenCodes.zip and look at the definitions tab, you can see it says "UA: Codes Unavailable for Assignment (e.g. N11, 555, 700, Home NPA, Test Codes, Set Aside Codes, etc.)".

There are two commits because I was not sure if negative lookaheads were allowed or not, feel free to merge if they are not, or discard my second commit if they are!


kkeshava:

Please check our issue tracker https://issuetracker.google.com/issues/75404127

PR #2537, opened 2020-10-20
Update demo to show reason for possibility fo all cases and show E.164 result even in not valid cases.

penmetsaa:
- The results/info like 'why it is possible' (Eg: IS_POSSIBLE_LOCAL_ONLY) is not displayed when generous IS_POSSIBLE is passed (which return true for local and full numbers).
- To be more informative and validity checks reasons, removing the conditional logic to display reason when only it is not possible.

This fixes:
b/171268264
b/27161576
More information there.


penmetsaa:
Pushed the changes into wrong origin. :) 
Please ignore this PR. Right one #2538 2538

PR #2538, opened 2020-10-20
JS and Java Demo updates to show isPossible Reason and E.164 for all cases

penmetsaa:
- The results/info like 'why it is possible' (Eg: IS_POSSIBLE_LOCAL_ONLY) is not displayed when generous IS_POSSIBLE is passed (which return true for local and full numbers).
- To be more informative and validity checks reasons, removing the conditional logic to display reason when only it is not possible.

This fixes:
- b/171268264
- b/27161576
More information there.


brittanyroesch (dismissed):

javascript/i18n/phonenumbers/demo.js
```diff
@@ -46,75 +46,82 @@ function phoneNumberParser() {
     output.append('****Parsing Result:****\n');
     output.append(goog.json.serialize(new goog.proto2.ObjectSerializer(
         goog.proto2.ObjectSerializer.KeyOption.NAME).serialize(number)));
-    output.append('\n\n****Validation Results:****');
+    output.append('\n\n****Possibility Results:****');
```
lararennie, line 49: Please do not rename this header. "Possibility results" doesn't make sense, and checking if a number is possible is a type of validation.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -46,75 +46,82 @@ function phoneNumberParser() {
...
+      case PNV.INVALID_COUNTRY_CODE:
+        output.append('INVALID_COUNTRY_CODE');
+        break;
+      case PNV.TOO_SHORT:
+        output.append('TOO_SHORT');
+        break;
+      case PNV.TOO_LONG:
+        output.append('TOO_LONG');
+        break;
+      case PNV.IS_POSSIBLE:
```
lararennie, line 65: I would break this up:
if possible, then say either nothing, or "N.B. LOCAL_ONLY" if that applies. If not possible, keep it as before but add the extra case statement for INVALID_LENGTH.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -46,75 +46,82 @@ function phoneNumberParser() {
...
+        break;
+      case PNV.IS_POSSIBLE_LOCAL_ONLY:
+        output.append('IS_POSSIBLE_LOCAL_ONLY');
+        break;
+      case PNV.INVALID_LENGTH:
+        output.append('INVALID_LENGTH');
+        break;
+    }
+    output.append('\nNote: numbers that are not possible have type ' +
+        'UNKNOWN, an unknown region, and are considered invalid.');
```
lararennie, line 76: This output was here so you did NOT have to then call isValidNumber for the non-possible case and print the stuff below. It makes no sense to say this and then print it.

Just change to if (isPossibleResult.equals(IS_POSSIBLE)) {
  // same code as before
}

javascript/i18n/phonenumbers/demo.js
penmetsaa: True. Gone with motion of product terminology.
Clubbed the results.

javascript/i18n/phonenumbers/demo.js
penmetsaa: Got it. It makes sense. As isValid() return false for local only number, should I stop displaying those? Like this:
https://screenshot.googleplex.com/BRXu6Lj3t2uuHwY
Not sure, I prefer displaying isValid() results also to provide more info as users may not have an idea about the hierarchy.

javascript/i18n/phonenumbers/demo.js
penmetsaa: True.. As I have reverted the change and having the check possible check, leaving as is.
Please review and suggest.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -50,10 +50,11 @@ function phoneNumberParser() {
     var isPossible = phoneUtil.isPossibleNumber(number);
     output.append('\nResult from isPossibleNumber(): ');
     output.append(isPossible);
+    var PNV = i18n.phonenumbers.PhoneNumberUtil.ValidationResult;
```
lararennie, line 53: PNV -> call it validationResult.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -63,57 +64,68 @@ function phoneNumberParser() {
         case PNV.TOO_LONG:
           output.append('TOO_LONG');
           break;
+        case PNV.INVALID_LENGTH:
+          output.append('INVALID_LENGTH');
+          break;
       }
       // IS_POSSIBLE shouldn't happen, since we only call this if _not_
```
lararennie, line 71: Expand comment to cover IS_POSSIBLE_LOCAL_ONLY too

javascript/i18n/phonenumbers/demo.js
```diff
@@ -63,57 +64,68 @@ function phoneNumberParser() {
...
       // possible.
       output.append('\nNote: numbers that are not possible have type ' +
           'UNKNOWN, an unknown region, and are considered invalid.');
     } else {
-      var isNumberValid = phoneUtil.isValidNumber(number);
-      output.append('\nResult from isValidNumber(): ');
-      output.append(isNumberValid);
-      if (isNumberValid && regionCode && regionCode != 'ZZ') {
+      // Checking as isValid() fails if possible local only.
+      if (isPossibleReason == PNV.IS_POSSIBLE_LOCAL_ONLY) {
```
lararennie, line 77: Like the idea: but won't this not trigger, because this is when isPossible is true??

penmetsaa: As isPossible returns true for possible national and local numbers this is getting triggered.
https://screenshot.googleplex.com/BRXu6Lj3t2uuHwY

lararennie: but it is not in the "if" block where isPossible is true? It looks to me that it is in the else block?

lararennie: (your screenshot is out-of-date, because it doesn't have the message your code does)

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done.

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done.

javascript/i18n/phonenumbers/demo.js
penmetsaa: It is ofcourse in the block where isPossible is true. This is else block of
if(!isPossible) {...}
else {}

lararennie: Oh dear :) I kept reading if (isPossible) :)
Can we actually revert the if and else block order to make it clear??.......... sorry!!

javascript/i18n/phonenumbers/demo.js
```diff
@@ -63,57 +64,68 @@ function phoneNumberParser() {
...
+        output.append('\nResult from isPossibleNumberWithReason(): ');
+        output.append('IS_POSSIBLE_LOCAL_ONLY');
+        output.append('\nNumber is considered as invalid as it is ' +
```
lararennie, line 81: remove "as" before invalid

javascript/i18n/phonenumbers/demo.js
```diff
@@ -63,57 +64,68 @@ function phoneNumberParser() {
...
+          'not a possible national number.');
+      } else {
+        var isNumberValid = phoneUtil.isValidNumber(number);
+        output.append('\nResult from isValidNumber(): ');
+        output.append(isNumberValid);
+        if (isNumberValid && regionCode && regionCode != 'ZZ') {
         output.append('\nResult from isValidNumberForRegion(): ');
```
lararennie, line 88: indentation looks off here

javascript/i18n/phonenumbers/demo.js
```diff
@@ -63,57 +64,68 @@ function phoneNumberParser() {
...
-          output.append('UAN');
-          break;
-        case PNT.UNKNOWN:
-          output.append('UNKNOWN');
-          break;
+        }
+        output.append('\nPhone Number region: ');
+        output.append(phoneUtil.getRegionCodeForNumber(number));
+        output.append('\nResult from getNumberType(): ');
+        var PNT = i18n.phonenumbers.PhoneNumberType;
```
lararennie: phoneNumberType

(also, can we put the switch statement below into a helper method, printType?)

lararennie (commented):
Ok, try and make it a bit more readable (the JS) and then update the Java to be the same.

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done. Made use of internal formatting tools. :)
Its painful without them

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done.

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done.

javascript/i18n/phonenumbers/demo.js
lararennie: You can set up vim to understand formatting rules, and then apply them to a file....

javascript/i18n/phonenumbers/demo.js
penmetsaa: Np :) Made the update so it is clear.

penmetsaa (commented):
Made the changes. Please review Lara.

penmetsaa:
Sorry. Pushed the changes lately because of hangover of process we follow for CL, commenting individually and then pushing. Sorry will make sure this do not happen.

Made changes to Java as well.
Corrected all the indenation issues, please let me know if I need to send them in different PR as diffs will be more.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -50,71 +62,52 @@ function phoneNumberParser() {
...
+        output.append('\nResult from isPossibleNumberWithReason(): ');
+        output.append('IS_POSSIBLE_LOCAL_ONLY');
+        output.append(
+            '\nNumber is considered invalid as it is ' +
+            'not a possible national number.');
+      } else {
+        var isNumberValid = phoneUtil.isValidNumber(number);
+        output.append('\nResult from isValidNumber(): ');
+        output.append(isNumberValid);
+        if (isNumberValid && regionCode && regionCode != 'ZZ') {
```
lararennie, line 79: we could add the !ZZ check in the Java as well for consistency.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -50,71 +62,52 @@ function phoneNumberParser() {
...
-          break;
-        case PNT.UAN:
-          output.append('UAN');
-          break;
-        case PNT.UNKNOWN:
-          output.append('UNKNOWN');
-          break;
-      }
+      output.append(
+          '\nNote: Numbers that are not possible or possible ' +
```
lararennie, line 108: same comment as per java

javascript/i18n/phonenumbers/demo.js
```diff
@@ -171,4 +164,32 @@ function phoneNumberParser() {
...
+      return 'PAGER';
+    case phoneNumberType.UAN:
+      return 'UAN';
+    case phoneNumberType.UNKNOWN:
+      return 'UNKNOWN';
+  }
+}
+
 goog.exportSymbol('phoneNumberParser', phoneNumberParser);
+
```
lararennie, line 195: remove blank line

lararennie (approved):

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done. In java the !ZZ check is present, it is covered in boolean variable initialization hasRegionCode. Looks like they did for readability. Changed, as both the versions LGTM.

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done.

javascript/i18n/phonenumbers/demo.js
penmetsaa: Done.

javascript/i18n/phonenumbers/demo.js
```diff
@@ -34,14 +34,26 @@ goog.require('i18n.phonenumbers.PhoneNumberUtil.ValidationResult');
 goog.require('i18n.phonenumbers.ShortNumberInfo');
 
 
+/**
+ * @const
+ * @private
+ */
+var phoneNumberType = i18n.phonenumbers.PhoneNumberType;
```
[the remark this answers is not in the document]
penmetsaa: Tried but receving the below error. We are actually using latest closure library and compiler. Sounds strnge.
I have replaced require statement with "const PhoneNumberType = goog.require('i18n.phonenumbers.PhoneNumberType');" and then removed earlier initalization statement. Please let me know if I am missing anything.

As we may need to change other initializations also, will check on this as a separate activity. WDYT?

     [exec] 
     [exec] /usr/local/google/home/penmetsaa/github_lab/libphonenumber/javascript/../../closure-library/closure/bin/build/closurebuilder.py: Compiling with the following command: java -client -jar /usr/local/google/home/penmetsaa/github_lab/libphonenumber/javascript/../../closure-compiler/bazel-bin/compiler_unshaded_deploy.jar --flagfile /tmp/tmpBS80th
     [exec] i18n/phonenumbers/demo.js:31:24: ERROR - [JSC_INVALID_CLOSURE_CALL_ERROR] Closure primitive methods (goog.provide, goog.require, goog.define, etc) must be called at file scope.
     [exec]   31| const phoneNumberType = goog.require('i18n.phonenumbers.PhoneNumberType');
     [exec]                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     [exec] 
     [exec] 1 error(s), 0 warning(s)
     [exec] Traceback (most recent call last):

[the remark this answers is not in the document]
lararennie: ah ok, thanks for trying.
I would tend to not do this const at all then, for consistency with everywhere else, and just say the whole string each time.

javascript/i18n/phonenumbers/demo.js
lararennie: Ahhhh thanks! I didn't go and check.
I'd go the other way and define hasDefaultRegion as a js var too, which is regionCode && regionCode != 'ZZ'.

Probably better.

penmetsaa: I also liked that apprach for readability :) 
Done.

javascript/i18n/phonenumbers/demo.js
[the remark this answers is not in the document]
penmetsaa: Done.

penmetsaa:
HI Lara,
Tested both JS and Java, working in all scenarios.

lararennie (approved):

PR #2548, opened 2020-11-25
Added fuzzer. 

DavidKorczynski:
This is first step in merging work from OSS-Fuzz into upstream, so we can remove the fuzzers from the OSS-Fuzz repository. 

I also checked the coverage we currently achieve, which is around 27% - so let me try in the near future at getting this a bit higher. 

Cross-referencing https://github.com/google/oss-fuzz/pull/4629

### DIFF
--- a/cpp/test/phonenumbers/fuzz_phone.cc
+++ b/cpp/test/phonenumbers/fuzz_phone.cc
@@ -1,0 +1,48 @@
+/* Copyright 2020 Google Inc.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+      http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


katbohm:
Makes sense, in principle it looks good to me. Some pieces seem to be missing (build files, dependencies, docs with operating instructions). Could you please add them?

DavidKorczynski:
So the fuzzing was reviewed by @penmetsaa in https://github.com/google/oss-fuzz/pull/4629 and is running on OSS-Fuzz since then. Would you like to migrate the fuzzers upstream or should I close this and leave as is?

katbohm:
Just to clarify -- the migration would be just adding this file upstream, but the fuzzing would still run on oss-fuzz or so, right? Or is something else needed for the migration?

penmetsaa:
Hi David,
Thanks much for these integration changes and they LGTM.
Good to know that we recognise any potential issues w.r.t invalid and fuzzy inputs.
All UTs are passing through at this commit..

DavidKorczynski:
> but the fuzzing would still run on oss-fuzz or so, right

yep that's correct! 

Fuzzer is removed from oss-fuzz here https://github.com/google/oss-fuzz/pull/8051

Thanks all!

PR #2556, opened 2020-12-16
Fix geocoding C++ build when static libraries are off.

Venemo:
I'm using the following CMake command to configure my libphonenumber build:

```
cmake -DCMAKE_BUILD_TYPE=Debug -DBUILD_GEOCODER=ON -DUSE_ALTERNATE_FORMATS=OFF -DUSE_ICU_REGEXP=ON -DUSE_LITE_METADATA=ON -DUSE_RE2=OFF -DBUILD_STATIC_LIB=OFF ..
```

This patch fixes the build with the above options. Without this patch CMake can't configure and build libphonenumber when geocoding is enabled but static library build is off.


gvollant:
@kevwongg-goog can pehaps take a loot, @penmetsaa

gvollant:
@Venemo I suggest you rebase against master

PR #2560, opened 2020-12-30
Fix building as dependency with add_subdirectory()

JoelLinn:
- When building as a dependency using cmake's add_subdirectory() use the
  current source directory instead of the top level source tree.

Previously it was not possible to use libphonenumber in such a way, especially when linking statically.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -175,7 +175,7 @@ if (${USE_STD_MAP} STREQUAL "OFF")
 endif ()
 
 # Add protoc (Protocol Buffers compiler) target.
-set (RESOURCES_DIR "${CMAKE_SOURCE_DIR}/../resources")
+set (RESOURCES_DIR "${CMAKE_CURRENT_SOURCE_DIR}/../resources")
 
 set (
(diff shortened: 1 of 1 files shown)


JoelLinn:
@googlebot I signed it!

JoelLinn:
Rebased.

JoelLinn:
I would be happy if anybody (maintainer or not) would be able to at least comment on this.

penmetsaa:
Thanks much for the change. Sorry for the delay here; because of some setup issues at my end, I could not test / look into it.
This change looks and sounds innocuous; and beneficiary w.r.t buidling from the end products points of view.
Followed regular CPP build process with this change and all tests passed.

penmetsaa (dismissed):

penmetsaa:
Please make the branch sync with master. Will merge after it.

gvollant (dismissed):

penmetsaa (dismissed):

penmetsaa:
Please sync the branch with master.
Out of date because of new merges.

penmetsaa (approved):

PR #2582, opened 2021-02-10
Update README.md

lionscribe:
Added info about the android port at https://github.com/lionscribe/libphonenumber-android which also supports geocoding, and only requires a one line code change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,7 +238,9 @@ Alternatives to our own versions:
     If you don't want to do this, check out the port at
     https://github.com/MichaelRocks/libphonenumber-android, which does repackage
     the metadata and use `AssetManager#open()`, and may be depended on without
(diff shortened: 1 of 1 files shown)


penmetsaa (approved):

penmetsaa:
Thanks so much for your contribution.
Did triage the option of contributing to the original android port (https://github.com/MichaelRocks/libphonenumber-android)?
This way we can reduce any redundancy and try have all flavours in one.

lionscribe:
MichaelRocks changed all the package names, which would make it incompatible as a quick drop-in replacement. An additional benefit of using the original package name, is that it can help a user prevent other libraries that they are importing from adding their own copy of Libphonenumber.

penmetsaa:
Thanks for looking into it. :)

PR #2620, opened 2021-05-17
Improve multithread perf of PhoneNumberUtil.isValidNumber

edrevo:
The synchronized access in `LRUCache` is a major performance bottleneck when using libphonenumber in multithreaded environments (in my case, I'm using this lib from Apache Spark).

A synthetic test that performs 10M phone number validations on 16 cores takes 30 seconds to complete. The proposed change uses a ThreadLocal cache to avoid synchronization, which reduces the runtime of the test to just 7.8 seconds (~x3.8 improvement).

I also prototyped an alternative solution that used a ReadWriteLock, but the perf improvement wasn't so good.


edrevo:
@googlebot I signed it!

PR #2621, opened 2021-05-19
Always use preferred intl prefix if present

penmetsaa:
After this change, the API always use preferred intl prefix if present, not just for numbers with a non-unique IDD. This means we will output "8~10" as the prefix if calling `formatOutOfCountryCallingNumber` instead of "810" for some regions that have this tilde in their prefix [designates that the user should wait before continuing to dial].

P.S: If `formatOutOfCountryCallingNumber()` API has returned format with wrong international exit code for region calling from, [please report the issue](https://github.com/google/libphonenumber/blob/master/CONTRIBUTING.md#filing-a-metadata-issue) as it has to be corrected in metadata.


pending_code_changes.txt
```diff
@@ -1 +1,2 @@
-
+Code changes:
+ - Always use preferred intl prefix if present, not just for numbers with a non-unique IDD. This means we will output "8~10" as the prefix if calling formatOutOfCountryCallingNumber instead of "810" for some regions that have this tilde in their prefix [designates that the user should wait before continuing to dial].
```
lararennie, line 2: I would phrase this as:
"Changes formatOutOfCountryCallingNumber to always use ....." so people understand the affected method more.

lararennie (approved):

lararennie (approved):

PR #2625, opened 2021-05-30
isRegionUsingNationalPrefix

Anrufliste:
Reducing the publication of metadata for a region code of https://github.com/google/libphonenumber/pull/2583 by just gaining access to its method hasNationalPrefix as an own method for a given RegionCode.

This method is directly used in the existing method getLengthOfGeographicalAreaCode where the beginning code is exactly the code needed for this functionality and thus can be replaced by it.

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -830,6 +830,19 @@ public class PhoneNumberUtil {
   }
 
   /**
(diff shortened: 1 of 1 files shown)


Anrufliste:
@googlebot I fixed it.

Anrufliste:
@googlebot I fixed it.

Anrufliste:
Hi, I think this is still a valid improvement and can be the base for some other improvements like https://issuetracker.google.com/issues/180311606 - can someone of the maintainer give me a feedback how the team see this pr? Thanks and Regards, Emil

PR #2640, opened 2021-07-19
Added Hormuud (25261) and moved 25268 to SomNet

byjamaljama:
* Updated 252.txt carrier metadata, adding SomNet (25268) which was previously (incorrectly) identified as NationLink
* Updated 252.txt carrier metadata, adding Hormuud (25261) which was previously missing from the list and as a result, the carrier mapper would result NULL for 25261XXXXXX numbers although it would still return the numbers valid.

For reference, see: https://en.wikipedia.org/wiki/Telephone_numbers_in_Somalia


byjamaljama:
@googlebot I signed it!

kkeshava:
Hi,

We have created a bug https://issuetracker.google.com/issues/195710760, please check it for further updates.

PR #2665, opened 2021-09-08
Update event list for workflow triggering.

penmetsaa:
- When configured "pull_request[type: opened] + push", workflow not getting triggered when there is a change in PR code/commits when the branch is from the parent's forked repo.
- Configured only pull_request event as any change can only go through PR. 
- pull_request event by default happens when opened, synchronize (means whenever there is change/commit, sync happens), and reopened. So this should be good enough for our use cases.

More details in [approach doc](https://docs.google.com/document/d/1YNYUAlEUXTi4UZG5gEFZGQ7oe7LxNKh7FYSg2qtIUvI/edit#heading=h.1c7qt5y1jb9)

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -19,7 +19,6 @@ jobs:
 # files. This is configured not to build the "cpp-build" module.
 # The "mvn install" command will run by default during the "install" phase by
 # Travis, without the profile flag. Here we customize the install phase to use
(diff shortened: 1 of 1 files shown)


penmetsaa:
This is to test whether Gitghub Actions are getting triggered from PR created through fork.
Inspired from: https://github.com/google/libphonenumber/pull/2627; may be that PR's upstream not yet updated with GA config.

gonecoding (approved):

PR #2674, opened 2021-09-15
Update FAQ about "why parsed national number is different in JS".

penmetsaa:
Inspired from https://issuetracker.google.com/issues/198423548
More details mentioned there.

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -534,3 +534,10 @@ This means we sometimes have numbers that do connect to a real person.
 If we by chance have actually listed your real number and would like it removed,
 please report this through Google's new [Issue Tracker](http://issuetracker.google.com/issues/new?component=192347).
 Only our internal team will have access to your identity (whereas GitHub usernames are public).
+
+### Why the national number parsed has different value than input in JavaScript?
+
+Eg: National number of 900184080594493**87**, ```region: JP``` is parsed as
+900184080594493**90**. Reason: When provided number is more than the max
+limit of JavaScript ```Number``` type - 2^53, JS starts rounding the value.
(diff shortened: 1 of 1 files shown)


gonecoding (approved):

FAQ.md
```diff
@@ -534,3 +534,10 @@ This means we sometimes have numbers that do connect to a real person.
 If we by chance have actually listed your real number and would like it removed,
 please report this through Google's new [Issue Tracker](http://issuetracker.google.com/issues/new?component=192347).
 Only our internal team will have access to your identity (whereas GitHub usernames are public).
+
+### Why the national number parsed has different value than input in JavaScript?
```
lararennie, line 538: Maybe better phrased as the following:
Why can the parsed national number have a different value in Javascript than what was entered?

lararennie: I think it's confusing though, because the parsed number can always have a different value. E.g. 642032345678, region NZ, would be 2032345678.

Ok a second try:

Why can the smallest digits of parsed numbers that are very long be incorrect when parsing in Javascript?

FAQ.md
```diff
@@ -534,3 +534,10 @@ This means we sometimes have numbers that do connect to a real person.
...
+
+Eg: National number of 900184080594493**87**, ```region: JP``` is parsed as
+900184080594493**90**. Reason: When provided number is more than the max
```
lararennie, line 541: Nit: Add "the" before "provided"

FAQ.md
penmetsaa: Yupp.. this is more accurate. Just mentioning national number may not be good enough for users.
Thanks Lara.

penmetsaa:
Hi Lara,

Made the updates. Could you please review and approve this?

lararennie (approved):

PR #2698, opened 2021-11-23
Be stricter about invalid UTF-8 input

neilmayhew:
Keep track of whether unicode strings were created from valid UTF-8 and use this to abort operations that use these strings later.

Typically, this is done by returning an empty string when a modified string is being produced, since an empty string is never a valid phone number.

See [b/206677455](https://issuetracker.google.com/issues/206677455)

---
This change is [Reviewable](https://reviewable.io/reviews/google/libphonenumber/2698)

### DIFF
--- a/cpp/src/phonenumbers/normalize_utf8.h
+++ b/cpp/src/phonenumbers/normalize_utf8.h
@@ -27,6 +27,8 @@ struct NormalizeUTF8 {
     string normalized;
     UnicodeText number_as_unicode;
     number_as_unicode.PointToUTF8(number.data(), static_cast<int>(number.size()));
+    if (!number_as_unicode.UTF8WasValid())
+      return normalized; // Return an empty result to indicate an error
     for (UnicodeText::const_iterator it = number_as_unicode.begin();
          it != number_as_unicode.end();
          ++it) {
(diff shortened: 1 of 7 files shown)


cpp/src/phonenumbers/normalize_utf8.h
```diff
@@ -27,6 +27,8 @@ struct NormalizeUTF8 {
     string normalized;
     UnicodeText number_as_unicode;
     number_as_unicode.PointToUTF8(number.data(), static_cast<int>(number.size()));
+    if (!number_as_unicode.UTF8WasValid())
+      return normalized;
```
katbohm, line 31: Maybe add a comment where normalized is declared (or here) to state that it is initialized to an empty string and this value is intended to be returned here. Also, do we have a test case covering that a non-utf8 string will be normalized to empty string?

neilmayhew: Done

penmetsaa:
Hi,

Thanks so much for the updates. The changes looks good for me.

Could you please update the branch with master merge.
So we can approve and merge?

penmetsaa:
Hi,
Could you please update the branch with master merge.
So we can approve and merge?

neilmayhew:
I rebased and squashed.

penmetsaa (approved):

neilmayhew:
I'm surprised you didn't mention this change in the release notes. However, it probably won't affect many people.

PR #2707, opened 2021-12-13
Disallow non-utf8 chars as input to phonenumbermatcher API

penmetsaa:
- Matcher.HasNext() returns false even in the first run when the input has non-utf8 chars in it.
-  Done by flagging such inputs when creating phonenumbermatcher obj.
- More details in b/201576863


cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -628,6 +627,7 @@ bool PhoneNumberMatcher::ExtractMatch(const string& candidate, int offset,
 bool PhoneNumberMatcher::HasNext() {
   if (state_ == NOT_READY) {
     PhoneNumberMatch temp_match;
+    try {
```
lararennie, line 630: For style reasons, we don't use exceptions in C++.

penmetsaa: Thanks Lara..
Sorry, earlier went ahead with the suggestion as we mistakenly assumed that we are not able to replicate internally and something to do with the flags like "-fsanitize=address,fuzzer-no-link" as we see lot memory leak warnings when run that way. We were able to reproduce the issue internally as well. cl/416205516

I see we hit this issue only when the input is large (> 2000 chars) and the index going negative  in next iteration. I updated the code to check the range of index. Will debug more and see the root cause and have proper fix (like having a cap on the input length / using unsigned int for index...

penmetsaa: Roughly till length 3500 we are not hitting the issue. As I am not familiar with CPP, could not debug why the index becomes negative. Maybe we can go ahead with the present solution?

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -656,6 +656,16 @@ bool PhoneNumberMatcher::Next(PhoneNumberMatch* match) {
 bool PhoneNumberMatcher::Find(int index, PhoneNumberMatch* match) {
   DCHECK(match);
 
+  // Input should should contain only UTF-8 characters.
```
lararennie, line 659: I don't think this is the right place to check this. That means you do this check every time you call next.
Can we do it in the constructor? The text does not change. And store it as a boolean, and if it's false, return false from HasNext?

penmetsaa: Hi Lara,

I have not blocked PhoneNumberMatcher creation as issues can be returned only through returning exceptions in constructor. As adding CPP exceptions are not encouraged (also introduce backward incompatibility -compile errors), lazily returning no matches (in Find() API), similar to earlier commits.

Yes, I would have explored alternate options like:
- doing this check in constructor and flag it to reference later during Find()/any as mentioned in https://stackoverflow.com/a/45643989
- I do not see other options that are innocuous. Please suggest.

cpp/test/phonenumbers/phonenumbermatcher_test.cc
```diff
@@ -1078,6 +1078,25 @@ TEST_F(PhoneNumberMatcherTest, NoMatchIfNoNumber) {
...
+  for (int i = 0; i < 100; ++i)
+    Numbers.append(stringWithSpecialCharacters);
+  scoped_ptr<PhoneNumberMatcher> matcher(
+      GetMatcherWithLeniency(Numbers, RegionCode::US(),
+                             PhoneNumberMatcher::POSSIBLE));
+  PhoneNumberMatch match;
+  while (matcher->HasNext()) {
+    matcher->Next(&match);
+  }
+  EXPECT_TRUE(true);
```
tvislavski, line 1097: EXPECT_TRUE(true)? :)

cpp/src/phonenumbers/phonenumbermatcher.cc
katbohm: Doing this check in constructor & adding a boolean flag sounds reasonable to me, to avoid checking the entire input text repeatedly.

cpp/test/phonenumbers/phonenumbermatcher_test.cc
penmetsaa: My bad. UT is not updated, checked only through debug results.
Updated now.

cpp/test/phonenumbers/phonenumbermatcher_test.cc
```diff
@@ -1090,11 +1090,15 @@ TEST_F(PhoneNumberMatcherTest, NoErrorWithSpecialCharacters) {
...
   while (matcher->HasNext()) {
-    matcher->Next(&match);
+    PhoneNumberMatch match;
+        matcher->Next(&match);
+    tries++;
   }
-  EXPECT_TRUE(true);
+  // As the input text do contain non UTF-8 chars, we block matching
+  // tries.
+  EXPECT_EQ(tries, 0);
```
tvislavski, line 1101: Can't you say here simply `EXPECT_FALSE(matcher->HasNext())`? Then you can omit `while` loop and counter

penmetsaa: +1. My bad.
I do not see reason now why I though it would be clear to mention that no tries will happen. :)

tvislavski (approved):
LGTM, but please wait for final approval from Lara or Katerina

katbohm (commented):
Do we document somewhere that the input for the matcher needs to be in clean utf8 and otherwise matcher.HasNext() returns false?

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: Done. Updated relevant .h file. Please verify whether it fine..

penmetsaa:
> 

Done. Updated relevant .h file. Please verify whether it fine..

katbohm (dismissed):
LGTM

cpp/src/phonenumbers/phonenumbermatcher.cc
lararennie: Indeed :)

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -1,4 +1,3 @@
-// Copyright (C) 2011 The Libphonenumber Authors
```
lararennie, line 1: Why did you drop the copyright?

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -407,7 +406,9 @@ PhoneNumberMatcher::PhoneNumberMatcher(const PhoneNumberUtil& util,
       max_tries_(max_tries),
       state_(NOT_READY),
       last_match_(NULL),
-      search_index_(0) {
+      search_index_(0),
+      isInputInCleanUTF8_(true) {
```
lararennie, line 410: clean -> valid?

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -421,12 +422,26 @@ PhoneNumberMatcher::PhoneNumberMatcher(const string& text,
       max_tries_(numeric_limits<int>::max()),
       state_(NOT_READY),
       last_match_(NULL),
-      search_index_(0) {
+      search_index_(0),
+      isInputInCleanUTF8_(true) {
+  isInputInCleanUTF8_ =  CheckInputUTF8OrNot();
 }
 
+
```
lararennie, line 430: Why the extra blank line?

cpp/src/phonenumbers/phonenumbermatcher.cc
lararennie: Also var names in c++ should be underscores, not camel case, so is_input_in_clean_utf8_

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -421,12 +422,26 @@ PhoneNumberMatcher::PhoneNumberMatcher(const string& text,
...
 PhoneNumberMatcher::~PhoneNumberMatcher() {
 }
 
+// static
+bool PhoneNumberMatcher::CheckInputUTF8OrNot() {
```
lararennie, line 435: Better method name is just IsInputUtf8 (rather than "or not" at the end)

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -421,12 +422,26 @@ PhoneNumberMatcher::PhoneNumberMatcher(const string& text,
...
+  // Input should contain only UTF-8 characters.
+  UnicodeText number_as_unicode;
+  number_as_unicode.PointToUTF8(text_.c_str(), text_.size());
+  if (number_as_unicode.UTF8WasValid()) {
```
lararennie, line 439: better: return number_as_unicode.UTF8WasValid(); instead of the if/else?

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -421,12 +422,26 @@ PhoneNumberMatcher::PhoneNumberMatcher(const string& text,
...
+      search_index_(0),
+      isInputInCleanUTF8_(true) {
+  isInputInCleanUTF8_ =  CheckInputUTF8OrNot();
 }
 
+
 PhoneNumberMatcher::~PhoneNumberMatcher() {
 }
 
+// static
```
lararennie, line 434: How/why is this static, doesn't it operate on "text_" which is a local var?

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -421,12 +422,26 @@ PhoneNumberMatcher::PhoneNumberMatcher(const string& text,
...
+bool PhoneNumberMatcher::CheckInputUTF8OrNot() {
+  // Input should contain only UTF-8 characters.
```
lararennie, line 436: Comment would be better if it explained *why* -> but then again, I wouldn't put it here, I would put it where you use the variable. Here the code doesn't need a comment IMO.

cpp/test/phonenumbers/phonenumbermatcher_test.cc
```diff
@@ -1078,6 +1078,23 @@ TEST_F(PhoneNumberMatcherTest, NoMatchIfNoNumber) {
...
+      "47\nfuzzvar1158: %415-666-1234 "
+      "i18n_phonenumbers_Pho\356eNumberMatcher_Leniency_VALID_1"
+      "\nfuzzvar1159: 20316 info:%415-666-7777 123 fake str79ee\nt";
+  string Numbers;
+  for (int i = 0; i < 100; ++i)
+    Numbers.append(stringWithSpecialCharacters);
+  scoped_ptr<PhoneNumberMatcher> matcher(
+      GetMatcherWithLeniency(Numbers, RegionCode::US(),
+                             PhoneNumberMatcher::POSSIBLE));
+  // As the input text do contain non UTF-8 chars, we block matching
```
lararennie, line 1093: Maybe a better comment: "Since the input text contains invalid UTF-8, we do not return any matches."

cpp/src/phonenumbers/phonenumbermatcher.h
```diff
@@ -101,6 +102,9 @@ class PhoneNumberMatcher {
     DONE,
   };
 
+  // API to check if the provided text_ is in UTF-8 or not.
```
lararennie, line 105: Remove "API to" and start with just "Checks if the ...."

cpp/src/phonenumbers/phonenumbermatcher.cc
lararennie: so perhaps: is_input_valid_utf8

cpp/src/phonenumbers/phonenumbermatcher.cc
```diff
@@ -626,6 +641,10 @@ bool PhoneNumberMatcher::ExtractMatch(const string& candidate, int offset,
 }
 
 bool PhoneNumberMatcher::HasNext() {
+  if (!isInputInCleanUTF8_) {
```
lararennie, line 644: This is where I would put a comment explaining *why* we early exit if it is not valid utf8 :)

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: +1. Done.

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: Done. +1

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: Removed.

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: Done. +1

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: My bad. Did by mistake.

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: Done. Thanks Lara :)

cpp/test/phonenumbers/phonenumbermatcher_test.cc
penmetsaa: Done.

cpp/src/phonenumbers/phonenumbermatcher.h
penmetsaa: Done. Thanks Lara :)

cpp/src/phonenumbers/phonenumbermatcher.cc
penmetsaa: Done. Thanks Lara :)

cpp/src/phonenumbers/phonenumbermatcher.h
lararennie: This one looks not done

lararennie (approved):

lararennie:
Update the PR description though too, it is a bit confusing right now.

penmetsaa:
> Update the PR description though too, it is a bit confusing right now.

Done. Thanks Lara :)

PR #2711, opened 2021-12-29
libphonernumber 25% speed improvement by SimpleItoa function

gvollant:
the function SimpleItoa create several temporary object.

This small modification use code from @jeaiii , project https://github.com/jeaiii/itoa , under MIT licence.

There is a 25% speed improvement without additionnal dependency or compatibility requirement

### DIFF
--- a/cpp/src/phonenumbers/stringutil.cc
+++ b/cpp/src/phonenumbers/stringutil.cc
@@ -14,10 +14,12 @@
 
 // Author: Philippe Liard
 
+#include <stdint.h>
 #include <algorithm>
 #include <cassert>
 #include <cstring>
 #include <sstream>
+#include <array>
 
 #include "phonenumbers/stringutil.h"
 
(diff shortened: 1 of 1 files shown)


gvollant:
Before merging, we need add acknoledgement to James Edward Anhalt III on CONTRIBUTORS and/or pehaps other files

gvollant:
@penmetsaa if you want speed improvement :-)

penmetsaa:
Sure.. We will triage/review on this.
However, metadata updates are the priority for us to ensure the data is fresh.
We will get to this as soon as we have some free time.

gvollant:
> Sure.. We will triage/review on this. However, metadata updates are the priority for us to ensure the data is fresh. We will get to this as soon as we have some free time.

the other option is code from https://github.com/google/libphonenumber/pull/2703#issuecomment-1002339155 , but we need option to compile libphone with c++17

katbohm:
Hi @gvollant, thank you for the proposed changes. However, as Aravind (#2703) mentioned, our intention is to go away from the String util implementations like this, replacing them with more functionality from absl packages. Thus, we currently won't be able to accept alternative implementations PRs like this one.

gvollant:
What is the solution to replace current SimpleItoa which take a lot of time?
I don't find an itoa like fast function in abeil

Pehaps the solution if offering option to compile with C++17 and use std::from_chars

katbohm:
Yes, replacing it with absl::StrCat is what I had in mind -- as you did in [#2731](https://github.com/google/libphonenumber/pull/2731) -- thank you for your contributions!

PR #2717, opened 2022-01-10
fixing: remove semicolon from the flags

katbohm:
fixing a reported bug: 
if (${USE_POSIX_THREAD} ... causes a syntax error, adding a semicolon to the flags
found out:
list (APPEND CMAKE_C_FLAGS ...) mistakenly adds a semicolon to the flags. 
set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread") solves the issue for me


penmetsaa:
LGTM..as:
- ```list (APPEND)``` is predominantly used to define list of libs and mentioning it when attaching sources to library.
- I see ```set ()``` command already been used in such cases of defining flags for execution.

penmetsaa:
I would like to hold my approval as today I am planning to do the release.
And we should be not merge to master branch, till its over.

penmetsaa (approved):

penmetsaa:
Could you please update the branch?
It needs, approving again. Will approve and merge.

penmetsaa (approved):

PR #2729, opened 2022-02-12
Falsehoods typo - prefix should be suffix

nicrodgers:
This particular falsehood talks about dialling more digits _after_ a number and so should be described as a suffix rather than prefix.

### DIFF
--- a/FALSEHOODS.md
+++ b/FALSEHOODS.md
@@ -120,7 +120,7 @@ surprising how many false assumptions programmers continue to make about them.
     after the country code (54) needs to be removed. This transforms +54 9 2982
     123456 into 02982 15 123456.
 
-1.  **No prefix of a valid phone number can be a valid phone number**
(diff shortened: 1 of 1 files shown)


netgsm:
Hi ,
There is now a new MVNO in Turkey. Can you add prefix 9051022 Netgsm Mobile

Nourhazimhmaidi (approved):

Nourhazimhmaidi (approved):

lararennie (changes requested):
Actually, the original was correct. The idea is basically that people assume since "999" is a number, that "999111" can't be, since you've already connected to 999 and the subsequent digits are just dropped/ignored. This is used for toll-free numbers in e.g. the US where it's 1800 LONGWORD and you could have dialled 1800 LONGWOR and it works too.

However, in a few countries, like Brazil, this does not hold, and you have to dial the exact numbers.

Perhaps we need to make the explanation more understandable?

PR #2731, opened 2022-02-18
Accelerate SimpleItoa conversion using abseil function

gvollant:
After reading comment from @katbohm and @penmetsaa on #2703 and #2711, I created this PR for faster itoa.

Using abseil function, SimpleItoa create less several temporary object than previous version and give speed improvement

### DIFF
--- a/cpp/src/phonenumbers/regexp_adapter_icu.cc
+++ b/cpp/src/phonenumbers/regexp_adapter_icu.cc
@@ -46,8 +46,7 @@ namespace {
 // Converts UnicodeString 'source' to a UTF8-formatted std::string.
 string UnicodeStringToUtf8String(const UnicodeString& source) {
   string data;
-  StringByteSink sink(&data);
-  source.toUTF8(sink);
+  source.toUTF8String(data);
   return data;
 }
 
--- a/cpp/src/phonenumbers/stringutil.cc
+++ b/cpp/src/phonenumbers/stringutil.cc
@@ -21,6 +21,9 @@
 
 #include "phonenumbers/stringutil.h"
 
+#include "absl/strings/str_cat.h"
+
+
 namespace i18n {
 namespace phonenumbers {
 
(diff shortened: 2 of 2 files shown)


gvollant:
@katbohm and @penmetsaa , do you known this "style" of modification correct with libphone rule?

regards

penmetsaa:
Hi @gvollant ,

Thanks so much for the understanding the status quo of maintaing an open source library :)
We appreciate your contributions for improving our CPP port.

This change looks good to me. I am in process of doing further improvements, mainly towards
use of abseil packages.
Still a draft version: https://github.com/google/libphonenumber/pull/2734

penmetsaa (approved):

gvollant:
Thank you.
I'm curious to known if you have speed enhancement with this modification

penmetsaa:
Hi @gvollant 
No we have not checked that..
We are just streamling the packages on what we depeneding on.
Is there any quick way that I can do benchmark testing?

gvollant:
One of my batch take 61 seconds instead 65 seconds.

I had more speed enhanchement with c++17 std::to_chars

penmetsaa:
Oh ok...
I think we are now operating at c++11 as min standard.
Maybe, it will be hard for other users as they need to recompile at c++17?
We generally discourage any forceful rebuild unless we have reasons; here,
as we mentioned earlier we are opting for consistency in using the utils.

Once my changes were done, maybe you help us by running the batch again.
I will check other team's CPP expert and conclude on approach, if performance
got hit badly.
P.S: We presently do not have CPP expert in team. Thanks for your help here.

gvollant:
And what about a choice between c++11 and c++17 as cmake parameter, then use std::to_chars is c++17 is detected?

penmetsaa:
Hi @gvollant ,
As of now, we cannot migrate to c++17 for above mentioned reasons.
Please let us know if you think otherwise.

gvollant:
I just suggest adding a "c++ version" option in the Cmake argument.
By default, we compile using c++11
If user select c++17, we can use c++17 feature

penmetsaa:
Hi,

Oh ok...
If the change do not force users to recompile at C++17, we are OK with it.
Please confirm my understanding: The api std::to_chars will direct to older implementations if C++17 is not present.
If yes, I have a PR in review, please help with a new PR after that: https://github.com/google/libphonenumber/pull/2734/files

P.S:
I am no CPP expert, please correct if I am wrong anywhere.
Please share your feedback on my PR. :)

PR #2734, opened 2022-02-23
More usage of abseil packages. Especially for string util usages.

penmetsaa:
- Faced link errors stemmed from mismatch of api definitions of absl::node_hash_set. So explicitly mentioned this lib as dependency during linking. This is common case/solution for absl hash set lib's link errors as per community.
- Decided to continued to have wrapper (stringutil), so that in future it's easy if we decided to change implementation; client code is safe. Changed most of the implementations & data structures in stringutil, to use abseil stuffs more.
- Though absl::string_view is suitable for most cases where we want immutable/read-only string structures, limiting to stringutil apis, not changing public API signatures (even though no compliance issue); reason: reduce dependency on abseil libs.
- Abseil specific changes done:
    - std::set => absl::node_hash_set
    - std::map => absl::node_hash_map
    - Using absl alternatives for “SplitStringUsing” API and “StringHolder” class impl.
    - Replaced manual impl of prefix and suffix checks with absl::StartsWith() and absl::EndsWith.
    - Similarly used absl::StrReplaceAll, absl::StrCat, absl::StrAppend where possible; and for type conversion apis.
- Note: Tried applying parameter packs to replace manually written duplicate ```StrCat``` apis (for accommodating more number of parameters of same type), but it did not go well when different type arguments passed. Eg char, int types to be converted to StringHolder; faced constructor implicit conversion issues.


cpp/src/phonenumbers/phonenumberutil.cc
```diff
@@ -2026,9 +2025,8 @@ bool PhoneNumberUtil::GetExampleNumberForType(
   DCHECK(number);
   std::set<string> regions;
   GetSupportedRegions(&regions);
-  for (std::set<string>::const_iterator it = regions.begin();
-       it != regions.end(); ++it) {
-    if (GetExampleNumberForType(*it, type, number)) {
+  for (string region_code : regions) {
```
katbohm, line 2028: const string& to avoid copying

cpp/src/phonenumbers/stringutil.cc
```diff
@@ -65,33 +65,12 @@ size_t FindNth(const string& s, char c, int n) {
...
-      result->push_back(substring);
-    }
-    start_pos = find_pos + delimiter.length();
-  }
-  if (start_pos != s.length()) {
-    result->push_back(s.substr(start_pos));
-  }
-}
-
-void StripString(string* s, const char* remove, char replacewith) {
```
katbohm, line 88: StripString already wasn't used anywhere, right?

cpp/src/phonenumbers/stringutil.cc
```diff
@@ -143,47 +117,20 @@ void strrmm(string* s, const string& chars) {
...
+StringHolder::StringHolder(const char* cp)
+  : absl::AlphaNum(cp)
 {}
 
 StringHolder::StringHolder(uint64 n)
-  : converted_string_(SimpleItoa(n)),
-    string_(&converted_string_),
-    cstring_(NULL),
-    len_(converted_string_.length())
+  : absl::AlphaNum(n)
```
katbohm, line 133: Note, if StringHolder gets a char, it might convert it in an undesired way. Possibly was true already before & maybe that never happens?

katbohm (commented):
In general LGTM, though we might want to check/test more that StringHolder is behaving as desired. 
I think that removing StrCat from stringutil and using absl::StrCat instead might simplify the code a bit more, so we might want to do that eventually (and removing stringutil altogether eventually possibly also).

cpp/src/phonenumbers/stringutil.cc
penmetsaa: Right.. Will test it with an UT just I am also curious here :) code: https://github.com/abseil/abseil-cpp/blob/master/absl/strings/str_cat.h#L263
As you mentioned, we do not have such a use case yet; so will take it offline.

cpp/src/phonenumbers/stringutil.cc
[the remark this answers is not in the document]
penmetsaa: Yes. :)

penmetsaa:
Yes, StrCat call could be direct to absl::StrCat. but idea is to not remove stringutil... the abstraction is good in opensource, so that we can switch the implementations without client code changes, clinet code Eg: PhoneNumberUtil... I see even more effective alternatives in new CPP17.. So leaving it easier to switch.

katbohm (approved):
LGTM

heftig:
I'm patching the build to use `find_package (absl REQUIRED)` instead of `FetchContent` to make it use a system `absl` package.

The produced shared lib doesn't work because it needs to link to `libabsl_raw_hash_set` but doesn't.

Shouldn't `absl::node_hash_set` be part of the `LIBRARY_DEPS`? Moving the dep from the test programs to the lib deps makes it work.

```diff
diff --git c/cpp/CMakeLists.txt i/cpp/CMakeLists.txt
index 588b97177..5aa450656 100644
--- c/cpp/CMakeLists.txt
+++ i/cpp/CMakeLists.txt
@@ -477,7 +477,7 @@ endif ()
 # Safeguarding against any potential link errors as mentioned in
 # https://github.com/abseil/abseil-cpp/issues/225
 set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
-list (APPEND LIBRARY_DEPS absl::strings)
+list (APPEND LIBRARY_DEPS absl::strings absl::node_hash_set)
 list (APPEND COMMON_DEPS absl::synchronization)
 
 if (APPLE)
@@ -570,7 +570,7 @@ if (NOT WIN32)
   list (APPEND TEST_LIBS pthread)
 endif ()
 
-target_link_libraries (libphonenumber_test ${TEST_LIBS} absl::node_hash_set)
+target_link_libraries (libphonenumber_test ${TEST_LIBS})
 
 # Unfortunately add_custom_target() can't accept a single command provided as a
 # list of commands.
@@ -677,7 +677,7 @@ if (${BUILD_GEOCODER} STREQUAL "ON")
     geocoding_test_program
     "test/phonenumbers/geocoding/geocoding_test_program.cc"
   )
-  target_link_libraries (geocoding_test_program geocoding phonenumber absl::node_hash_set)
+  target_link_libraries (geocoding_test_program geocoding phonenumber)
 endif ()
 
 # Build an RPM
```

gvollant:
for information , here is the list of static library I used to link my app which use libphone
```
libphonenumber.a libicui18n.a libicuuc.a libicudata.a libprotobuf-lite.a libabsl_hash.a libabsl_strings.a libabsl_low_level_hash.a libabsl_city.a libabsl_raw_hash_set.a libabsl_raw_logging_internal.a libabsl_throw_delegate.a
```

I want produce a Linux (and Windows) executable and dynamic library for my application which contain all libphone and dependency in the binary. So user don't need install library or several file

penmetsaa:
Did some cleanup in https://github.com/google/libphonenumber/pull/2753
Please contribute if more makes sense.

PR #2743, opened 2022-03-15
Using CPP11 'auto' instead of iterators

penmetsaa:
Inspired from cl/434264801
If there are any such replaceable code snippets (s/iterators/auto), will do in my next cleanup.


katbohm (approved):
LGTM

PR #2744, opened 2022-03-15
Making it clear that safe_strtou64 kind of string to int conversion apis fail when non decimal chars are present.

penmetsaa:
Fix for UT breakage after https://github.com/google/libphonenumber/pull/2742.
```
[ RUN      ] StringUtilTest.safe_strtou64
/usr/local/google/home/penmetsaa/github_lab/libphonenumber/cpp/test/phonenumbers/stringutil_test.cc:139: Failure
Expected equality of these values:
  18446744073709551615ULL
    Which is: 18446744073709551615
  n
    Which is: 0
```
Though there is a failure in conversion, fortunately the job is done, though absl mark it as [unspecified state](https://github.com/abseil/abseil-cpp/blob/5ed77665c4d697f657fa1362b5a482450f858cdc/absl/strings/numbers.h#L63) 
Looks like just "void" is better sol in this case. More details there.

### DIFF
--- a/cpp/src/phonenumbers/stringutil.cc
+++ b/cpp/src/phonenumbers/stringutil.cc
@@ -88,7 +88,8 @@ bool HasSuffixString(const string& s, const string& suffix) {
 
 template <typename T>
 void GenericAtoi(const string& s, T* out) {
-  (void)absl::SimpleAtoi(s, out);
+  if (!absl::SimpleAtoi(s, out))
+    *out = 0;
 }
 
 void safe_strto32(const string& s, int32 *n) {
--- a/cpp/src/phonenumbers/stringutil.cc
+++ b/cpp/src/phonenumbers/stringutil.cc
@@ -88,8 +88,7 @@ bool HasSuffixString(const string& s, const string& suffix) {
 
 template <typename T>
 void GenericAtoi(const string& s, T* out) {
-  if (!absl::SimpleAtoi(s, out))
-    *out = 0;
+  (void)absl::SimpleAtoi(s, out);
 }
 
 void safe_strto32(const string& s, int32 *n) {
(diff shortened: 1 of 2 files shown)


penmetsaa:
b/224732913 to make github actions (Out externa CI/CD pipeline) test CPP UTs as well..

gvollant:
with "void" we rely on "unspecified" state.
In "real" libphone code, is it possible we hav "UL" or "ULL" suffix?

katbohm (commented):
What is safe_strtou64 supposed to do? Get a number for a string that is a valid number, or also for a string that can contain some noise? 
In case it's supposed to work only for valid numbers, then the test doesn't seem right:
  safe_strtou64("18446744073709551615UL", &n);
  EXPECT_EQ(18446744073709551615ULL, n);

In case it's supposed to work even if there is noise at the end, absl:simpleAtoi might leave it in an unspecified state (see below copy pasted from here: https://github.com/abseil/abseil-cpp/blob/master/absl/strings/numbers.h):
// SimpleAtoi()
//
// Converts the given string (optionally followed or preceded by ASCII
// whitespace) into an integer value, returning `true` if successful. The string
// must reflect a base-10 integer whose value falls within the range of the
// integer type (optionally preceded by a `+` or `-`). If any errors are
// encountered, this function returns `false`, leaving `out` in an unspecified
// state.

penmetsaa:
Sorry, I just went through the expectations of test.
Yes its not righteous to rely on unspecified state; checked the usages, the expectation is clear that it has to convert decimal number string to int. Checked scenario of calls like:
safe_strtou64(normalized_national_number, &number_as_int);
safe_strtou64(standard_rate_example, &national_number);

Fixed the test now than func.

gvollant (approved):

katbohm (approved):
LGTM (but maybe change the name of the PR?)
Thanks!

penmetsaa:
Done. Thanks for both of you for guiding in right direction. :)

PR #2749, opened 2022-04-03
libphonenumber must be linked with absl::node_hash_set

AdrianBunk:
It is used in the library with #2734, and not linking with it
breaks users of the library: https://bugs.debian.org/1008817


penmetsaa:
Hi,

We were build successfully and UT all the CPP binaries/targets. We were building in:
Operating System: Debian GNU/Linux rodete
Kernel: Linux 5.15.15-1rodete2-amd64

As this change inspires some cleanups, approving same.

penmetsaa (approved):

PR #2752, opened 2022-04-11
France - Allow 0692 mobile numbers from new allocated ranges

Wrongusername:
Range allocation decision
https://www.arcep.fr/uploads/tx_gsavis/21-0150.pdf


rohininidhi:
Hi,

As per ITU: https://www.itu.int/dms_pub/itu-t/oth/02/02/T020200004B0002PDFE.pdf, prefix 692 is not assigned to France (+33) instead belongs to Reunion with country code +262.

Please input the number using the correct country code i.e +262 692012345 or you can provide defaultCountry as 'RE'. This range is already marked by our library as a valid MOBILE number, see the demo: [http://libphonenumber.appspot.com/phonenumberparser?number=692012345&country=RE]

PR #2753, opened 2022-04-11
Removing duplicate / explicit linking of absl packages to lib binaries.

penmetsaa:
Inspirted from: https://github.com/google/libphonenumber/pull/2749.
Mostly just a code convienience, no build / runtime changes.


katbohm (approved):
LGTM

PR #2763, opened 2022-04-24
Make abseil-cpp a git submodule

darkdragon-001:
Building in environments without network access (e.g. Flatpak, reproducible builds, ...) it is problematic to download sources from CMake. Following recommendations from the [abseil documentation](https://github.com/abseil/abseil-cpp/blob/master/CMake/README.md#step-by-step-instructions) of using a git submodule.

Related issue: https://issuetracker.google.com/issues/230166349


.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "tools/cpp/abseil-cpp"]
+	path = tools/cpp/abseil-cpp
```
penmetsaa, line 2: Maybe "cpp/deps" as this is the main source directory.
As we are touching cpp build system in existing PR https://github.com/google/libphonenumber/pull/2818/files, we would like to review that first.

tools/cpp/CMakeLists.txt
```diff
@@ -28,30 +28,10 @@ project (generate_geocoding_data)
...
-     message (FATAL_ERROR "Could not build abseil-cpp binaries.")
-  endif ()
-
-  # Safeguarding against any potential link errors as mentioned in
-  # https://github.com/abseil/abseil-cpp/issues/225
-  set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
-  add_subdirectory(${abseil-cpp_SOURCE_DIR} ${abseil-cpp_BINARY_DIR})
-endif()
+# Safeguarding against any potential link errors as mentioned in
+# https://github.com/abseil/abseil-cpp/issues/225
```
penmetsaa, line 32: For product that uses abseil present/installed in the system, having the reference as a git module may not be pointing to right version..  Eventhough there is a way to configure version, it may not be the exact one used i.e no validations happening. How do we solve such cases? I do not have much experience in this area.

penmetsaa:
Hi,

I am not much aware of the git submodules;
We may take more time to review this as another PR https://github.com/google/libphonenumber/pull/2818/files is in review process.

tools/cpp/CMakeLists.txt
darkdragon-001: We could also put it into a `if(NOT absl_FOUND)` such that the submodule is only used when it is not present in the system.

PR #2793, opened 2022-07-19
Fixating the absl-cpp version that we download and cpp readme update after #2791

penmetsaa:
- We are cloning only a particular version of the abseil-cpp so that we
   are not commiting to catch up with new build requirements of that
   library. Eg: In recent absl-cop release C++14 is made as minimum
   version to build the library.
- **Error details in b/239501609**
- Readme update w.r.t new abseil-cpp dependency changes i.e after
   #2791

### DIFF
--- a/cpp/README
+++ b/cpp/README
@@ -135,6 +135,31 @@ Requirements:
     Note: Boost Thread is the only library needed at link time.
 
 
+  - abseil-cpp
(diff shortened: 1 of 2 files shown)


katbohm (dismissed):
LGTM

katbohm (approved):
LGTM

PR #2818, opened 2022-09-09
CPP: Fix cmake build failure + export libphonenumber-config.cmake

temideoye:
Currently, the project fails to build when `BUILD_GEOCODER=OFF` due to a missing dependency. The issue is described [here](https://github.com/google/libphonenumber/pull/2791#issuecomment-1236621218). This push fixes the bug and proposes the following additional features to enhance the CMake build process overall:

1. Expose `BUILD_SHARED_LIBS` and `BUILD_TESTING` as user configurable options
2. Add functionality to build and export CMake package config files so downstream projects can consume the library with a simple `find_package(libphonenumber)`, without creating a FindLibPhonenumber.cmake file
3. Issue: https://issuetracker.google.com/issues/239501100


temideoye:
Hey, @penmetsaa! Just checking in, have you been able to review this?

penmetsaa:
Hi @temideoye ,

Thanks so much for your patience. We are planning to review this in next 4 weeks.

cpp/CMakeLists.txt
```diff
@@ -16,10 +16,13 @@
 
 cmake_minimum_required (VERSION 3.11)
 
+project (libphonenumber VERSION 8.12.0)
```
penmetsaa, line 19: Remove the version from here.

penmetsaa, line 19: Could you please change the version to 8.13.0?

cpp/CMakeLists.txt
```diff
@@ -16,10 +16,13 @@
...
+
 # Pick the C++ standard to compile with.
 # Abseil currently supports C++11, C++14, and C++17.
 set(CMAKE_CXX_STANDARD 11 CACHE STRING "C++ standard used to compile this project")
 set(CMAKE_CXX_STANDARD_REQUIRED ON)
+set(CMAKE_POSITION_INDEPENDENT_CODE TRUE)
 
 project (libphonenumber)
```
penmetsaa, line 27: Please remove this line as project is already defined; and without version we are getting error: 
```
CMake Error at /usr/share/cmake-3.24/Modules/WriteBasicConfigVersionFile.cmake:43 (message):
  No VERSION specified for WRITE_BASIC_CONFIG_VERSION_FILE()
Call Stack (most recent call first):
  /usr/share/cmake-3.24/Modules/CMakePackageConfigHelpers.cmake:237 (write_basic_config_version_file)
  CMakeLists.txt:706 (write_basic_package_version_file)
```

tools/cpp/CMakeLists.txt
```diff
@@ -23,37 +23,6 @@ set(CMAKE_CXX_STANDARD_REQUIRED ON)
 
 project (generate_geocoding_data)
 
-# Helper functions dealing with finding libraries and programs this library
```
penmetsaa, line 26: As we have moved the download functionality to main library's CMAKE, we were unable to build the offline geocoder individually. Could we not remove this from here OR have this as a common build rule file?

```
CMake Error at CMakeLists.txt:39 (target_link_libraries):
  Target "generate_geocoding_data" links to:

    absl::strings

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.
```
Note: I am not expertise in CPP builds, please suggest and do what you think is best :)

cpp/CMakeLists.txt
```diff
@@ -83,25 +86,54 @@ function (find_required_program NAME FILENAME DESCRIPTION)
...
+option (BUILD_GEOCODER "Build the offline phone number geocoder" ON)
+option (REGENERATE_METADATA "Regenerate metadata instead of using it from the source tree" ON)
+option (USE_ALTERNATE_FORMATS "Use alternate formats" ON)
+option (USE_PROTOBUF_LITE "Link to protobuf-lite" OFF)
+option (USE_BOOST "Use Boost" ON)
+option (USE_ICU_REGEXP "Use ICU regexp engine" ON)
+option (USE_LITE_METADATA "Use lite metadata" OFF)
+option (USE_RE2 "Use RE2" OFF)
+option (USE_STD_MAP "Force the use of std::map" OFF)
+option (BUILD_STATIC_LIB "Build static libraries" ON)
```
penmetsaa, line 98: About these new options, could you please add some notes in cpp/README file.
It helps readers to know about these options.

penmetsaa (changes requested):
Thanks so much for meticulous work.
Mentioned some minor suggestions / doubts I have, please take a look. :)

cpp/CMakeLists.txt
temideoye: Fixed with [3ce80e8](https://github.com/google/libphonenumber/pull/2818/commits/3ce80e80368daaadf3b315e2d4398eb52c7b7978).

temideoye: By the way, I removed calls to `set(libphonenumber_VERSION_MAJOR 8)` and co. They are unnecessary; CMake automatically sets the variables on call to `project(libphonenumber VERSION 8.13.0)`.

tools/cpp/CMakeLists.txt
penmetsaa: Can we do anything about this..?

penmetsaa:
Thanks +1 to what you mentioned.. Some pending AIs:
- My comment on: tools/cpp/CMakeLists.txt
- Documentation update (a new PR) regarding the new oprions

penmetsaa (approved):

temideoye:
@penmetsaa, that looks like cmake couldn't find your abseil lib directory. Do you have the installation path of abseil in [CMAKE_PREFIX_PATH](https://cmake.org/cmake/help/latest/variable/CMAKE_PREFIX_PATH.html)? You may try something like `cmake -S "path/to/libphonenumber/cpp" -B "path/to/build/directory" -DCMAKE_PREFIX_PATH="path/to/abseil;path/to/other/dependencies" `.

tools/cpp/CMakeLists.txt
temideoye: @penmetsaa, I had assumed (erroneously) that tools/cpp is merely a dependency of the main project, but now I understand you build it independently for other purposes as well.

There are 2, or 3 possible ways to resolve this:

1. Copy the find abseil functions back to tools/cpp. The demerit of this is obviously the maintenance overhead.
2. Create a new shared cmake module for the find abseil functions. This is DRYer than 1, but still a bit of a dance.
3. Build tools/cpp via the main project with an option like `-DBUILD_TOOLS_ONLY=ON`. This should simplify the interface in that all dependency resolutions happen in the same file. It would be my preference too.

I will share a pull request in a moment. Let me know if it works for you, or if you'd rather resolve it a different way.

penmetsaa:
Hi @temideoye 

- When I tried installing abseil at system level and then mentioning path in the CMAKE_PREFIX_PATH -> it is working correctly.
- When there is no abseil found, we know that FetchContent() dowloads and builds the abseil to temp/cache dir and referring from there -> This is working before our/this commit.

Is it that after the change, the cached abseil binaries are not been able to found?

temideoye:
Oh dang, I see what you mean. The issue seems to be that abseil disables its install rules in superbuilds:

https://github.com/abseil/abseil-cpp/blob/b0cc11b9763d50f6ecd160719b0f78fde9712a7f/CMakeLists.txt#L60-L66

Setting `ABSL_ENABLE_INSTALL=ON` should fix the errors. My bad, I should have also tested the build without a prebuilt abseil. Will send you a patch to test in a moment.

temideoye:
@penmetsaa, I've added a fix for the install errors to [2869](https://github.com/google/libphonenumber/pull/2869). Let me know if it resolves the issue for you.

penmetsaa:
Thanks much for the patch @temideoye..
Closing the thread as that fix resolved the issue

cpp/CMakeLists.txt
```diff
@@ -358,7 +382,7 @@ add_metadata_gen_target (
   ${TEST_METADATA_TARGET}
   "${RESOURCES_DIR}/PhoneNumberMetadataForTesting.xml"
   "test_metadata"
-  "test_metadata"
+  "metadata"
```
nekopsykose, line 385: this rename conflicts with the metadata.h generated above in the non-test METADATA_TARGET. they both output a `metadata.h` like this, which means the build has two rules that generate the same output file (one for metadata_target, and one for test_metadata_target).

this breaks the build for cmake -G Ninja, because the .ninja spec forbids two distinct rules for the same output:  

```console
$ ninja -C build/
ninja: entering directory 'build/'
ninja: multiple rules generate '/home/demon/src/aports/community/libphonenumber/src/libphonenumber-8.13.4/cpp/src/phonenumbers/metadata.h'
```

temideoye: @nekopsykose, thanks for the catch. The bug was fixed [here](https://github.com/google/libphonenumber/commit/2369340a05076ef235be1550f202ff5b66db9917), but was reintroduced by an clumsy merge. @penmetsaa, will send in a quick pull to restore the fix if that's OK.

temideoye: Regression fixed with [2874](https://github.com/google/libphonenumber/pull/2874).

penmetsaa:
Thanks @nekopsykose  and @temideoye Merged the #2874 PR

nekopsykose:
@penmetsaa thanks!

PR #2834, opened 2022-10-12
Get resource fix

roggnroll:


roggnroll:
doesn't work, it get MalformedURLException

penmetsaa:
MalformedURLException is resolved after adding absolute paths from proj home (WEB-INF/....); works locally, however still facing issue after deploying in live app engine server. Same error.
```
java.lang.NullPointerException
at com.google.template.soy.SoyFileSet$Builder.add(SoyFileSet.java:365)
```
Tried configuring static files path in appengine web.xml. Need to check further

PR #2839, opened 2022-10-18
Using precompiled templates with SoySauce

tvislavski:
Using precompiled templates with SoySauce instead of Tofu runtime interpreter

### DIFF
--- a/java/demo/src/main/java/com/google/phonenumbers/demo/render/LibPhoneNumberRenderer.java
+++ b/java/demo/src/main/java/com/google/phonenumbers/demo/render/LibPhoneNumberRenderer.java
@@ -26,7 +26,7 @@ public abstract class LibPhoneNumberRenderer<T extends SoyTemplate> {
 
   String render(T soyTemplate) {
     SoySauce soySauce = new SoySauceBuilder().build();
-    return soySauce.newRenderer(soyTemplate).renderText().get();
+    return soySauce.newRenderer(soyTemplate).renderHtml().get().toString();
   }
 
   public abstract String genHtml();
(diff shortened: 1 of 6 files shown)


java/demo/src/main/java/com/google/phonenumbers/demo/render/LibPhoneNumberRenderer.java
```diff
@@ -18,32 +18,15 @@
...
   String render(T soyTemplate) {
-    SoyFileSet sfs = SoyFileSet.builder().add(getResource()).build();
-    SoyTofu tofu = sfs.compileToTofu();
-    // For convenience, create another SoyTofu object that has a
-    // namespace specified, so you can pass partial template names to
-    // the newRenderer() method.
-    SoyTofu simpleTofu = tofu.forNamespace(namespace);
-    return simpleTofu.newRenderer(soyTemplate).render();
+    SoySauce soySauce = new SoySauceBuilder().build();
+    return soySauce.newRenderer(soyTemplate).renderText().get();
```
roggnroll, line 29:
```suggestion
    return soySauce.newRenderer(soyTemplate).renderHtml().get().toString();
```

roggnroll (changes requested):
Excellent job! I have only a small change request

java/demo/src/main/java/com/google/phonenumbers/demo/render/LibPhoneNumberRenderer.java
roggnroll: For text templates, renderer.renderText().get() will return a String instead of a SanitizedContent object, since the text is NOT sanitized.

roggnroll (approved):
Superb work, thanks!

PR #2854, opened 2022-11-23
Penmetsaa maven update for release 8.13.1

penmetsaa:
Along with release version changes, correcting the maven lifecycle phase in demo's pom,
so that soy classes are compiled before the artifact is built.


rohininidhi (approved):

penmetsaa:
The isssue is fixed in: https://github.com/google/libphonenumber/pull/2855

PR #2860, opened 2022-12-07
Add missing test_metadata.h

i386x:
Without `test_metadata.h`, when running with `REGENERATE_METADATA=OFF`, `cmake` deletes `test_metadata.c` and then fails with `file not found` error.


penmetsaa:
Hi,

Thanks so much for the contributions here.. The changes looks good to me.
Maybe because the metadata is not exposed externally, we may have not written an header for it.
Do you think we are missing any such .h file creation? - irrespective of we having public apis in them or not..

PR #2868, opened 2022-12-31
Move find_package call for absl to global scope

cordlandwehr:
absl is needed for compilation also with GEOCODER disabled.

Signed-off-by: Andreas Cord-Landwehr <cordlandwehr@kde.org>


penmetsaa:
Hi,

Thanks so much for making this change..
We have a PR under review for same fix and more improvements in CPP build process.
https://github.com/google/libphonenumber/pull/2818/
So closing this PR for now.

PR #2869, opened 2023-01-05
CPP: Add BUILD_TOOLS_ONLY option

temideoye:
- The project currently fails to find abseil and gtest while building` libphonenumber/tools/cpp` because dependency resolutions have been consolidated in `libphonenumber/cpp`. This commit proposes a BUILD_TOOLS_ONLY option be added to the main project to allow building tools/cpp without duplicating common dependencies.
- In essence, to build libphonenumber/tools/cpp, do `cmake -S "path/to/libphonenumber/cpp" -DBUILD_TOOLS_ONLY=ON`
- This also solves the issue of ["Absl not being able to Install in super builds"](https://github.com/google/libphonenumber/pull/2818#issuecomment-1371867829)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -109,6 +109,9 @@ endif ()
 find_package(absl)
 
 if(NOT absl_FOUND)
+  # Overide abseil install rules for subprojects
+  set(ABSL_ENABLE_INSTALL ON)
+  
   # Downloading the abseil sources at particular version to not catch up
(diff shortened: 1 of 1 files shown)


penmetsaa (approved):

penmetsaa:
Thanks much for the patch @temideoye ..
It definately helps us in fixing the build issue.
We will definately let you know if you we would still like trigger the build of Build tools separately.
Mostly not needed as far as I know.

PR #2883, opened 2023-02-06
Solving leaks within Windows

trishvl:
There's a bug where the critical section used by in the win32 Singleton implementation is never deleted. This causes it to be leaked within Windows. Therefore, changing it to a different type of lock, SRWLOCK which doesn't require "init" or "deinit" and there will be nothing to leak. Also removing the double-check for (once_init_) since with the SRWLOCK, it is no longer needed.

There's also another bug within the phonenumberutil.cc file where the PhoneMetadata members are being leaked. This PR adds clean up code for that.

### DIFF
--- a/cpp/src/phonenumbers/base/memory/singleton_win32.h
+++ b/cpp/src/phonenumbers/base/memory/singleton_win32.h
@@ -30,14 +30,12 @@ class Singleton {
   virtual ~Singleton() {}
 
   static T* GetInstance() {
+    AcquireSRWLockExclusive(&singleton_lock_);
     if (once_init_) {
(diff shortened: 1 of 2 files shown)


aschoorl (approved):

trishvl:
@gvollant Hi Gilles! Considering that you've worked on adding support of singleton and lock for win32 and c++ 11 here in libphonenumber, I was wondering if you'd be able to take a look at this PR. Thank you - Trish

Muhammadbobur3 (approved):

trishvl:
Hoping to ping contributors and get more eyes on this, thanks! @penmetsaa @rohininidhi @temideoye @hagbard

PR #2889, opened 2023-02-15
Syncing all open-source files to github.

hagbard:
This includes previously unsynced code for regular expression manipulation and the finite state matcher.

PiperOrigin-RevId: 472261412


hagbard:
I'm just making sure the metadata code is up-to-date in open source. It was never pushed after the regex/matcher code got moved there. This includes a few library updates in existing code which happened in the meantime (it's good to push this periodically).

The default PR misidentified me, so someone will need to force the CLA.

penmetsaa (approved):

PR #2920, opened 2023-04-27
Update ShortNumberMetadata.xml

Ken-UC:
could you add 3646 3949 as national short number


kkeshava:
Sorry for the delay and Thank you for contributing to libphonenumber!

Both the shortcodes are supported in our Metadata, so closing this as WAI.

PR #3030, opened 2023-05-26
 Source: Refactor and reformat code using clang-format and clang-tidy 

zenon8adams:
This commit includes several changes to improve the readability and maintainability of the codebase. These changes include:

    Replacing raw-loops with for-each loops and standard algorithms where possible
    Rewriting duplicate StrAppend and StrCat functions with variadic versions
    Using ArraySize to calculate the size of arrays at compile-time instead of using C-style methods
    Annotating inheriting methods with the override keyword
    Using automatic type deduction where appropriate
    Using emplace_back instead of push_back where possible
    Replacing string.compare method with direct equality comparison on strings
    Defaulting trivial constructors and destructors
    Replacing NULL with nullptr


cpp/src/phonenumbers/asyoutypeformatter.cc
```diff
@@ -89,31 +86,19 @@ PhoneMetadata CreateEmptyMetadata() {
...
-      current_output_(),
-      formatting_template_(),
-      current_formatting_pattern_(),
-      accrued_input_(),
-      accrued_input_without_formatting_(),
-      able_to_format_(true),
-      input_has_formatting_(false),
-      is_complete_number_(false),
-      is_expecting_country_code_(false),
+    : regexp_factory_(new RegExpFactory()), regexp_cache_(*regexp_factory_, 64),
```
roubert, line 89: I don't think that joining such lines as these here increase readability or maintainability, on the contrary, joining these lines will make any future changes to these lists require a reformatting. That's not an improvement.

cpp/src/phonenumbers/base/basictypes.h
```diff
@@ -5,13 +5,13 @@
 #ifndef I18N_PHONENUMBERS_BASE_BASICTYPES_H_
 #define I18N_PHONENUMBERS_BASE_BASICTYPES_H_
 
-#include <limits.h>         // So we can set the bounds of our types
-#include <stddef.h>         // For size_t
-#include <string.h>         // for memcpy
+#include <climits>  // So we can set the bounds of our types
```
roubert, line 8: For such files as this one, that are copied verbatim from other open source projects, I think it would be better to not touch them at all, for it makes maintenance easier if they're unchanged so that newer versions can be easily copied from their projects of origin.

zenon8adams: Are this the only changes I should leave? I'll like to know the files this applies to. Thanks.

roubert: > Are this the only changes I should leave?

I don't know. Sorry.

zenon8adams: I'll go through the files and keep legacy changes. Thanks.

PR #3297, opened 2023-11-10
Use a singleton for ShortNumberInfo in PhoneExtractor::RunFocusedArena

svasudevan91:
This change is expected to reduce CPU usage by only initializing ShortNumberInfo once throughout the lifetime of the application. This pattern is already used in PhoneNumberUtil.

### DIFF
--- a/cpp/src/phonenumbers/shortnumberinfo.h
+++ b/cpp/src/phonenumbers/shortnumberinfo.h
@@ -44,9 +44,6 @@ class PhoneNumberUtil;
 
 class ShortNumberInfo {
  public:
-  ShortNumberInfo();
-  ~ShortNumberInfo();
-
   // Cost categories of short numbers.
   enum ShortNumberCost {
     TOLL_FREE,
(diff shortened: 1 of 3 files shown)


cpp/test/phonenumbers/shortnumberinfo_test.cc
```diff
@@ -42,12 +42,12 @@ class ShortNumberInfoTest : public testing::Test {
     return phone_number;
   }
 
-  ShortNumberInfoTest() : short_info_() {
-    PhoneNumberUtil::GetInstance()->SetLogger(new StdoutLogger());
+  ShortNumberInfoTest() : short_info_(ShortNumberInfo::GetInstance()) {
+   PhoneNumberUtil::GetInstance()->SetLogger(new StdoutLogger());
```
katbohm, line 46: I think the space should stay.

cpp/src/phonenumbers/shortnumberinfo.h
```diff
@@ -44,9 +44,6 @@ class PhoneNumberUtil;
 
 class ShortNumberInfo {
  public:
-  ShortNumberInfo();
-  ~ShortNumberInfo();
```
katbohm, line 48: By removing these public constructor & destructor, the public API is changing. I don't think we can do that since the library is widely used.

katbohm (commented):
Hi there! Thank you for your pull request. My main note here is that by removing the public constructor and replacing it by getInstance() the public API is changed. Since LibPhoneNumber is a widely used library we shouldn't just remove an existing API.

PR #3298, opened 2023-11-14
Added two methods to get mobile number length of any country

WalidAhmed90:
There was no method to get the mobile number length from Country ISO Code. So, I created couple of functions to achieve that. Below is the example

    public int getPhoneNumberLength(String countryCode) {
        int countryCodeInt = Integer.parseInt(countryCode);
        String isoCode = getRegionCodeForCountryCode(countryCodeInt);
        PhoneNumber examplePhoneNumber = getExampleNumber(isoCode, PhoneNumberType.MOBILE);

        if (examplePhoneNumber != null) {
            String exampleNumber = examplePhoneNumber.getNationalNumber();
            return exampleNumber.length();
        } else {
            return 0; // No example Mobile number available
        }
    }

and I overloaded one method

    public PhoneNumber getExampleNumber(String regionCode, PhoneNumberType phoneNumberType) {
        return getExampleNumberForType(regionCode, phoneNumberType);
    }

Just use it like this:

    /**
     * Example of how to use the getPhoneNumberLength method to calculate the length
     * of a valid national Mobile number for a given country code.
     *
     * Usage:
     * ```java
     * public int getPhoneNumberLength(String countryCode) {
     *     PhoneNumberUtil phoneNumberUtil = PhoneNumberUtil.getInstance();
     *     return phoneNumberUtil.getPhoneNumberLength(countryCode);
     * }
     * ```
     *
     * @param countryCode The numeric country code for which the example Mobile number length is needed.
     *                    Example: "92" for Pakistan, "971" for the United Arab Emirates (UAE).
     * @return The length of a valid national Mobile number for the specified country code.
     *         Returns 0 if the country code is invalid or if no example Mobile number is available.
     *         Returns -1 if the metadata does not contain such information.
     */
    public int getPhoneNumberLength(String countryCode) {
        PhoneNumberUtil phoneNumberUtil = PhoneNumberUtil.getInstance();
        return phoneNumberUtil.getPhoneNumberLength(countryCode);
    }

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -27,6 +27,7 @@ import com.google.i18n.phonenumbers.internal.RegexCache;
 import com.google.i18n.phonenumbers.metadata.DefaultMetadataDependenciesProvider;
 import com.google.i18n.phonenumbers.metadata.source.MetadataSource;
 import com.google.i18n.phonenumbers.metadata.source.MetadataSourceImpl;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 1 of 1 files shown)


WalidAhmed90:
...

PR #3300, opened 2023-11-21
Avoid unnecessary copies of PhoneMetadata

sopwithcamel:
The PhoneMetadata can be moved into the hashmap avoiding a copy.


katbohm (approved):
LGTM

mandlil (approved):

PR #3301, opened 2023-11-21
fix: update CMakeLists.txt to allow building with REGENERATE_METADATA off

CoolGoose:
I am not sure if this is the correct change (since my experience with Cmake is limited) but from my understanding and errors I am getting the current else branch breaks the current generated metadata becasue of the OUTPUT flag.

The "fix" is to just display the skip message in the else, and it proceeds to build without any issues.


CoolGoose:
I am not sure what's wrong with the CLA but it errors out with

`Your CLA signature could not be read; please reach out to your contact Googler.`

rfeiner:
Any update about this one? 🙏

katbohm (approved):
LGTM

PR #3357, opened 2023-12-14
Replace usage of deprecated Callback4 with std::function.

Cory-Kramer:
No changes needed in other client languages, since this is a private implementation detail.


katbohm (approved):
Thank you!

PR #3359, opened 2023-12-27
Hash pin arduino/setup-protoc version

joycebrum:
Hi, I'm from GOSST. I've noticed that just the arduino/setup-protoc was not hash pinned on the codeql workflow. Here is a PR to hash pin it, keeping it on the 1st major version.

Thanks!


tvislavski (approved):
Thank you!

mandlil (approved):

PR #3623, opened 2024-09-11
Generate RegionCode constants for testing.

silvio2402:
This generates the RegionCode class that contains string constants for testing using the build tools.

Context: b/237181793


tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -195,6 +205,16 @@ public boolean start() {
 
       writeCountryCallingCodeMappingToJavaFile(
           countryCodeToRegionCodeMap, outputDir, mappingClass, copyright);
+      
+      if (buildRegioncode) {
+        SortedSet<String> regionCodeSet = new TreeSet<String>();
+        regionCodeSet.addAll(getDefaultRegionCodeList());
+        regionCodeSet.addAll(BuildMetadataFromXml.buildRegionCodeList(metadataCollection));
+        List<String> regionCodeList = new ArrayList<String>(regionCodeSet);
```
tvislavski, line 213: There is no need to wrap this inside a list, you can just pass a set to the method. You are using a for loop to iterate over it, and since it is a sorted set, the iteration will happen in exactly the same way.

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -339,6 +359,32 @@ private static void writeCountryCodeSet(ClassWriter writer, int capacity,
     writer.addToBody("  }\n");
   }
 
+  private static final String REGION_CODE_CONSTS_COMMENT =
```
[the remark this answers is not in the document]
silvio2402: Makes sense. But then we should also move the [other constants](google/libphonenumber/blob/master/tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java#L222-L234) in this file someday.

java/libphonenumber/test/com/google/i18n/phonenumbers/RegionCode.java
```diff
@@ -14,51 +14,49 @@
...
+ * Please don't modify it directly.
+ */
+
 package com.google.i18n.phonenumbers;
 
-/**
- * Class containing string constants of region codes for easier testing.
- */
 final class RegionCode {
+  // Class containing string constants of region codes for easier testing.
```
tvislavski, line 24: Sorry, I just now noticed that this is not a Javadoc comment anymore. Could you also change the `ClassWriter` to accept a Javadoc on top of the class? Also consider a Builder pattern for the `ClassWriter`, as you would have now different combinations of constructor parameters. You could use [AutoValue.Builder](https://github.com/google/auto/blob/main/value/userguide/builders.md) to have it autogenerated.

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -382,7 +428,7 @@ void writeToFile() throws IOException {
         writer.write("\n");
       }
 
-      writer.write("public class " + name + " {\n");
+      writer.write((modifiers.isEmpty() ? "" : modifiers + " ") + "class " + name + " {\n");
```
tvislavski, line 431: `modifiers` could also be null in this case

tools/java/java-build/src/com/google/i18n/phonenumbers/BuildMetadataProtoFromXml.java
```diff
@@ -376,60 +391,89 @@ private static void writeRegionCodeConstantsToJavaFile(Collection<String> region
...
+        importsBuilder().add(name);
+        return this;
+      }
 
-    void addToBody(CharSequence text) {
-      body.append(text);
-    }
+      abstract Builder setJavadoc(String javadoc);
+      private final StringBuilder javadocBuilder = new StringBuilder();
+      final Builder addToJavadoc(String text) {
```
tvislavski, line 428: Do you actually need this? `setJavadoc` might be enough.

tvislavski (dismissed):

mandlil (dismissed):

tvislavski (approved):

mandlil (approved):

PR #3627, opened 2024-09-18
Rename NonGeographicalRegion to NonGeographicalEntity.

KarolJakubKrawiec:
This is to be consistent with the naming of non-geographical entities, since region and non-geographical can be contradictory.

Context: b/206752240


tvislavski (commented):
Can you also update https://github.com/google/libphonenumber/blob/master/pending_code_changes.txt with the explanation of this change? This will be included in the release notes then

tvislavski (dismissed):

cpp/src/phonenumbers/phonenumberutil.h
```diff
@@ -865,6 +865,9 @@ class PhoneNumberUtil : public Singleton<PhoneNumberUtil> {
   const i18n::phonenumbers::PhoneMetadata* GetMetadataForNonGeographicalRegion(
       int country_calling_code) const;
```
silvio2402, line 866: You can remove this since it is a private function.

cpp/test/phonenumbers/phonenumberutil_test.cc
```diff
@@ -66,7 +66,12 @@ class PhoneNumberUtilTest : public testing::Test {
 
   const PhoneMetadata* GetMetadataForNonGeographicalRegion(
       int country_code) const {
-    return phone_util_.GetMetadataForNonGeographicalRegion(country_code);
+    return phone_util_.GetMetadataForNonGeographicalEntity(country_code);
+  }
```
silvio2402, line 70: You can also remove this because you already renamed the calls to this function.

silvio2402 (approved):
LGTM

mandlil (approved):

PR #3769, opened 2024-12-13
Add "()" when string.match() is executed in js

SG-Kang:
Without "()", when many conditions exist and former one is matched, it returns the wrong value. So adding "()" for the regex

I think it doesn't happen in Java and cpp

[related issue] (raised by myself)
https://issuetracker.google.com/issues/383824582

[java]
https://github.com/google/libphonenumber/blob/master/java/libphonenumber/src/com/google/i18n/phonenumbers/internal/RegexBasedMatcher.java#L49-L56

[cpp]
https://github.com/google/libphonenumber/blob/master/cpp/src/phonenumbers/regex_based_matcher.cc#L51-L65

### DIFF
--- a/javascript/i18n/phonenumbers/phonenumberutil.js
+++ b/javascript/i18n/phonenumbers/phonenumberutil.js
@@ -4760,7 +4760,7 @@ i18n.phonenumbers.PhoneNumberUtil.prototype.canBeInternationallyDialled =
 i18n.phonenumbers.PhoneNumberUtil.matchesEntirely = function(regex, str) {
   /** @type {Array.<string>} */
   var matchedGroups = (typeof regex == 'string') ?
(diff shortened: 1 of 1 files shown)


SG-Kang:
> Hi,
> 
> Thank you for contributing to libphonenumber!
> 
> We tested with your PR changes, internally we are getting testcase failures for the string type phonenumbers. Instead of writing changes in matchEntirely function, you can write here(
> 
> https://github.com/google/libphonenumber/blob/424617599369e7adba8a5d1509b910d9ce2e3e44/javascript/i18n/phonenumbers/phonenumberutil.js#L3855
> 
> ) like **new RegExp('^(?:' + generalDesc.getNationalNumberPatternOrDefault() + ')$')**.

mandlil (approved):

PR #3780, opened 2025-01-10
Refactor phone context parsing for RFC3966 numbers.

silvio2402:
- Refactor the logic to extract the phone context from an RFC3966 URI to a new `PhoneContextParser` class.
- Refactor the number normalization methods `normalizeDigits` and `normalizeDigitsOnly` to a new `PhoneNumberNormalizer` class.
  > This keeps `PhoneNumberUtil.normalizeDigitsOnly` as a shell because it is public.
- Move the constants that are common between `PhoneContextParser` and `PhoneNumberUtil` to a new `Constants` class.

This is a refactor and should not change the logic of the parser. 

Context: b/370922064


silvio2402:
Refer to new PR: https://github.com/google/libphonenumber/pull/3791

PR #3791, opened 2025-01-22
Refactor phone context parsing for RFC3966 numbers.

silvio2402:
- Refactor the logic to extract the phone context from an RFC3966 URI to a new `PhoneContextParser` class.
- Refactor the number normalization methods `normalizeDigits` and `normalizeDigitsOnly` to a new `PhoneNumberNormalizer` class.
  > This keeps `PhoneNumberUtil.normalizeDigitsOnly` as a shell because it is public.
- Move the constants that are common between `PhoneContextParser` and `PhoneNumberUtil` to a new `Constants` class.

This is a refactor and should not change the logic of the parser. 

Context: b/370922064


java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneContextParser.java
```diff
@@ -0,0 +1,166 @@
...
+          "The phone-context value is invalid.");
+    }
+
+    return parsePhoneContext(phoneContext);
+  }
+
+  /** Represents the parsed phone-context parameter of an RFC3966 tel-URI. */
+  static class PhoneContext {
+    /** The raw value of the phone-context parameter. */
+    private String rawContext_ = null;
```
tvislavski, line 132: Please remove trailing underscores after field names

java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneContextParser.java
```diff
@@ -0,0 +1,166 @@
...
+    if (phoneContext.charAt(0) != Constants.PLUS_SIGN) {
+      return new PhoneContext().setRawContext(phoneContext).setCountryCode(null);
+    }
+
+    // Remove the plus sign from the phone context and normalize the digits.
+    String normalizedPhoneContext =
+        PhoneNumberNormalizer.normalizeDigitsOnly(phoneContext.substring(1));
+
+    // Check if the phone context is a valid country calling code.
+    if (!normalizedPhoneContext.equals("")
```
tvislavski, line 95: How about introducing Constants.MIN_LENGTH_COUNTRY_CODE

tvislavski (dismissed):

PR #3952, opened 2026-01-06
Remove outdated paragraph about Kosovo's country code

ArditBaloku:
As of Februrary 2017, Kosovo has it's own country code of `+383` making this paragraph in the FALSEHOODS.md documentation incorrect. The call will fail if the codes 381, 386, or 377 are used.
https://en.wikipedia.org/wiki/Telephone_numbers_in_Kosovo

### DIFF
--- a/FALSEHOODS.md
+++ b/FALSEHOODS.md
@@ -58,16 +58,6 @@ surprising how many false assumptions programmers continue to make about them.
     The USA, Canada, and several Caribbean islands share the country calling
     code +1. Russia and Kazakhstan share +7. These are not the only examples!
 
-1.  **Each country has only one country calling code**
-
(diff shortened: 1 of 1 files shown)


kkeshava:
Sorry for the delayed response and Thank you for taking the time to improve libphonenumber!

We understand that Kosovo has its own country code but we still support the ranges under both countries and the same is mentioned in the XML file
+377 - https://github.com/google/libphonenumber/blob/b950178dea75cd87c2242b01fc6225a4d729c005/resources/PhoneNumberMetadata.xml#L20159
+386 - https://github.com/google/libphonenumber/blob/b950178dea75cd87c2242b01fc6225a4d729c005/resources/PhoneNumberMetadata.xml#L27032

We will investigate it in https://issuetracker.google.com/issues/520292514, please check it for further updates.

PR #3987, opened 2026-04-15
Fix ReDoS in RFC3966 domain validation and integer underflow in matcher

sharadboni:
## Summary

- **Bug A (ReDoS):** The `rfc3966_domainlabel_` and `rfc3966_toplabel_` regex patterns in `phonenumberutil.cc` use nested quantifiers `[chars]+((\\-)*[chars])*` that cause O(N^2) backtracking on adversarial inputs via ICU regex. This is reachable through the `phone-context` parameter in RFC3966 URIs parsed by `IsPhoneContextValid()`. The fix rewrites both patterns to `[chars]+(-[chars]+)*`, which requires at least one character after each hyphen, eliminating the quantifier nesting ambiguity.

- **Bug B (Integer underflow):** In `AllNumberGroupsAreExactlyPresent()` in `phonenumbermatcher.cc`, when `candidate_groups` is empty, `candidate_groups.size() - 1` underflows `size_t` (unsigned) producing a huge value. This is then cast to `int` and used as a vector index, causing undefined behavior. The fix adds an early `return false` guard when `candidate_groups` is empty.

## Test plan

- [ ] Verify existing C++ unit tests pass (the regex change is semantically equivalent for valid RFC3966 domain names since hyphens in DNS labels must be followed by at least one alphanumeric character)
- [ ] Verify that parsing a phone number with a long crafted `phone-context` domain no longer causes quadratic regex time
- [ ] Verify that `AllNumberGroupsAreExactlyPresent` returns false for empty candidate groups instead of triggering UB

### DIFF
--- a/cpp/src/phonenumbers/phonenumbermatcher.cc
+++ b/cpp/src/phonenumbers/phonenumbermatcher.cc
@@ -837,6 +837,11 @@ bool PhoneNumberMatcher::AllNumberGroupsAreExactlyPresent(
     candidate_groups.push_back(digit_block);
   }
 
+  // Guard against size_t underflow when candidate_groups is empty.
+  if (candidate_groups.empty()) {
+    return false;
+  }
+
   // Set this to the last group, skipping it if the number has an extension.
   int candidate_number_group_index = static_cast<int>(
       phone_number.has_extension() ? candidate_groups.size() - 2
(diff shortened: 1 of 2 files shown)


sharadboni:
@rohininidhi Could you review this security fix? It rewrites ReDoS-vulnerable regex patterns in RFC3966 domain validation and fixes an integer underflow in AllNumberGroupsAreExactlyPresent.

PR #4016, opened 2026-06-20
Introduce MetadataBytes wrapper for compiled-in metadata accessors

hjanuschka:
## Summary

Introduces a small `MetadataBytes` RAII wrapper for the compiled-in metadata
buffers returned by the internal `metadata.h`, `short_metadata.h` and
`alternate_format.h` accessors.

The C-style API

```cpp
int metadata_size();
const void* metadata_get();
```

is replaced by

```cpp
MetadataBytes GetMetadata();   // and GetShortMetadata(), GetAlternateFormat()
```

where `MetadataBytes` is a move-only handle that holds either:

1. a non-owning `const void*` + `int size`, or
2. a `std::unique_ptr<uint8_t[]>` + `int size` (owning).

The default implementations shipped in this repo continue to wrap the
existing `static const unsigned char data[]` arrays with a non-owning
instance, so on-disk layout and runtime behaviour are unchanged for
upstream consumers. `CppMetadataGenerator` (and its tests) are updated to
emit the new API.

## Backwards compatibility

These headers are library-internal (not part of the public API).

## Related

Chromium CL discussion that motivated this:
https://chromium-review.googlesource.com/c/chromium/src/+/7928952


battre (approved):
LGTM!

kkeshava (dismissed):

mandlil:
Hi @hjanuschka, 
It looks like the C++ build is failing with these changes. Could you please take a look and update the PR?

hjanuschka:
@mandlil tests pass now!

katbohm (approved):
LGTM

mandlil (approved):

PR #4018, opened 2026-06-30
Include <utility> for std::move in metadata_bytes.h

hjanuschka:
metadata_bytes.h uses std::move in the owning constructor but only includes `<cstdint>` and `<memory>`. It builds fine when `<utility>` happens to come in transitively, but breaks under a libc++ modules build (hit this rolling into Chromium) where std::move has to be imported from its own header. Adding the include fixes it.


battre (approved):
LGTM

hjanuschka:
cc @mandlil

mandlil (approved):

katbohm (approved):

PR #4070, opened 2026-08-27
Avoid catastrophic backtracking in the RFC3966 domainname check

shaggyinsomniac:
The per-label pattern used to validate the `phone-context` domain during parsing is ambiguous:

```
alnum+((\\-)*[alnum])*
```

An alphanumeric run can be split between the leading `+` and the repeating group in many equivalent ways, so when the overall match fails, a backtracking engine re-explores the suffix once per split — exponential in the number of labels.

Java's own matcher prunes this shape quickly (~1 ms on a 123-char failing input on JDK 26), so the JVM library isn't practically affected. However:

- the **JS port** mirrors this pattern and hangs for over **2 minutes** on a 93-char failing `phone-context` (V8),
- the **Python port** (daviddrysdale/python-phonenumbers) mirrors it too and takes **~50 s** on 98 chars through its public `parse()` (CPython 3.10, phonenumbers 9.0.37).

Repro for the JS shape:

```js
const DOMAINNAME = /^([a-zA-Z0-9]+((\\-)*[a-zA-Z0-9])*\\.)*[a-zA-Z]+((\\-)*[a-zA-Z0-9])*\\.?$/;
const t = Date.now();
DOMAINNAME.test("aa.".repeat(30) + "aa!");
console.log(Date.now() - t);   // ~147000 ms
```

This rewrites the label as `alnum+(\\-+[alnum]+)*` — same accepted language (alnum runs separated by one or more hyphens, still no leading/trailing hyphen or empty labels), but each position matches exactly one way. With the rewrite the repro above is 0 ms.

Changes `PhoneNumberUtil.java` and the hand-maintained JS mirror in `phonenumberutil.js`, and adds tests pinning the accepted language (single and repeated hyphens accepted; leading/trailing hyphens and empty labels rejected) plus rejection of the long ambiguous phone-context. Full `mvn -pl java/libphonenumber test` suite passes locally (285 tests, 0 failures).

Same fix proposed in daviddrysdale/python-phonenumbers#324.

This was reported to Google's security team, who assessed it as below the bar for security tracking and suggested public disclosure.

### DIFF
--- a/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
+++ b/java/libphonenumber/src/com/google/i18n/phonenumbers/PhoneNumberUtil.java
@@ -328,9 +328,9 @@ public class PhoneNumberUtil {
   // defined in RFC3966.
   private static final String ALPHANUM = VALID_ALPHA + DIGITS;
   private static final String RFC3966_DOMAINLABEL =
-      "[" + ALPHANUM + "]+((\\-)*[" + ALPHANUM + "])*";
+      "[" + ALPHANUM + "]+(\\-+[" + ALPHANUM + "]+)*";
   private static final String RFC3966_TOPLABEL =
-      "[" + VALID_ALPHA + "]+((\\-)*[" + ALPHANUM + "])*";
+      "[" + VALID_ALPHA + "]+(\\-+[" + ALPHANUM + "]+)*";
   private static final String RFC3966_DOMAINNAME =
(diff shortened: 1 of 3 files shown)


mandlil:
Thanks for the fix!

The proposed `RFC3966_TOPLABEL` pattern rejects valid domains containing digits like `example.c1` and `example.b2b`.
Please add test cases and update `RFC3966_TOPLABEL` (in Java and JS) to:
```java
"[" + VALID_ALPHA + "][" + ALPHANUM + "]*(\\-+[" + ALPHANUM + "]+)*"

